PR #25, opened 2015-07-20
CLIO-23417 Add text input component

jnraine:


seanhealy:
Closing as @JefCurtis took over this a while ago and we left this hanging by accident.

PR #28, opened 2015-08-19
clio-24393 fix arrow centering

seanhealy:
Also, fixed a popover init error that would occasionally crop up on the
docs viewer.

### DIFF
--- a/themis_components/thPopover/index.scss
+++ b/themis_components/thPopover/index.scss
@@ -31,8 +31,7 @@ i.th-popover-arrow {
     display: block;
     position: absolute;
     margin-top: 12px;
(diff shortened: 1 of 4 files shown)


themis_components/thPopover/index.scss
```diff
@@ -57,7 +56,7 @@ i.th-popover-arrow {
   left: 0px;
   right: 0px;
 
-  background-color: rgba(0,0,0,0.1);
+  background-color: rgba(0,0,0,0);
```
seanhealy: We do want to eat that click. After opening a menu / popover / modal the next click is a dismiss action. You can see that for example in OS X menus.

PR #32, opened 2015-08-25
Clio 24731 thdropdown

JefCurtis:


themis_components/thDropdown/readme.md
```diff
@@ -0,0 +1,35 @@
+# Dropdown Button — `thDropdown`
+
+### Description
+
+Add a button that when clicked displays a dropdown list.
+
+`thDropdown` will accept a array of object that when created correctly will display each as list items
```
seanhealy: **an** array

themis_theme/utility_classes.scss
```diff
@@ -0,0 +1,41 @@
+/*--------------------------------------------------
+* Utility classes and general styles
```
seanhealy: What is this for?

themis_theme/reset.scss
```diff
@@ -0,0 +1,48 @@
+/* http://meyerweb.com/eric/tools/css/reset/
+   v2.0 | 20110126
```
seanhealy: Do we need a reset? Browsers have normalized a lot since they were popular.

seanhealy:
We initially talked also about allowing for the items to be able to be specified from markup. Did you decide this isn't a good idea. I was thinking this would allow the component to be a bit more accessible for simple cases when you may not want / need a controller. 

```html
<th-dropdown name="Button Name">
  <th-item icon="foo" name="Item 1" />
  <th-item icon="bar" name="Item 2" />
</th-dropdown>
```

JefCurtis:
@seanhealy This is something that we can still do but the feedback was that it was probably a good idea to push work into Angular controllers. Get everybody into the habit. That is certainly something I can add.

seanhealy:
As we discussed yesterday with the select I think we should do the `<th-item>` here as well. It's trivial to add and makes this a whole lot more useful as I picture there being a number times you'd want to use this without a controller.

themis_components/thDropdown/examples/3 - with th-item/coffee.coffee
```diff
@@ -0,0 +1,7 @@
+angular.module('thDemo', ['ThemisComponents'])
+  .controller "DemoController", ->
+    @links = [
```
seanhealy: I don't think we need an example showing off mixing a controller list and markup list. It might just be confusing. Perhaps take the list out of this one…

themis_components/thDropdown/thDropdown.directive.coffee
```diff
@@ -0,0 +1,104 @@
...
+      ng-click="dropdown.toggle()">
+      {{dropdown.name}}
+      <i class="fa" ng-class="dropdown.toggleCaret()"></i>
+    </a>
+    <div
+      ng-click="dropdown.toggle()"
+      ng-if="dropdown.visible"
+      class="dropdown-menu"
+      >
+      <span ng-repeat-start="item in dropdown.processedItems"></span>
```
seanhealy: Lets use ng-switch here as we discussed.
http://codepen.io/anon/pen/BoKJXd

themis_components/thDropdown/thDropdown.directive.coffee
```diff
@@ -0,0 +1,104 @@
...
+        <th-item
+          ng-if="item.type == 'link'"
+          name="{{item.name}}"
+          url="{{item.url}}"
```
seanhealy: `href`

themis_components/thDropdown/thDropdown.directive.coffee
```diff
@@ -0,0 +1,104 @@
...
+          icon="{{item.icon}}"
+        >
+        </th-item>
+        <th-item
+          ng-if="item.type == 'action'"
+          name="{{item.name}}"
+          action="item.action"
```
seanhealy: `ng-click`

PR #37, opened 2015-09-02
Clio 23419 thSelect Design docs

JefCurtis:


themis_components/thSelect/readme.md
```diff
@@ -0,0 +1,156 @@
+# DESIGN DOCS
```
seanhealy: We should probably put this in the `design.md` file. :wink:

themis_components/thSelect/readme.md
```diff
@@ -0,0 +1,156 @@
...
+</div>
+```
+![Angular](http://zippy.gfycat.com/CornyNegativeCorydorascatfish.gif "Angular Select")
+
+
+## THOUGHTS
+---
+- `<optgroup>` should be as easy as just passings separated named arrays
+  ```
+  groceries = [
```
seanhealy: This syntax isn't valid. We probably want to do something along the lines of:

```js
groceries = [
  {
    name: "Dairy"
    contents: [
      { name: "Eggs", value: "eggs" }
      { name: "Milk", value: "milk" }
    ]
  }
  {
    name: "Vegetable"
    contents: [
      { name: "Lettuce", value: "lettuce" }
      { name: "Carrots", value: "carrots" }
    ]
  }
  {
    name: "Fruit"
    contents: [
      { name: "Apple", value: "apple" }
      { name: "Cherries", value: "cherries" }
    ]
  }
]
```

themis_components/thSelect/readme.md
```diff
@@ -0,0 +1,156 @@
...
+    ]
+    fruit = [
+      { name: "Apple", value: "apple" }
+      { name: "Cherries", value: "cherries" }
+    ]
+  ]
+  ```
+- Creating a wrapper angular component for Select2 would solve a lot of our problems and could get us moving in the right direction quickly.
+- We could roll out the different options as they are ready but allowing them to be passed in as options.
+- If this library is meant to be open source we are not adding dependencies in jQuery and Select2. I'm not sure if that is important or not.
```
seanhealy: In this line do you mean that `Select2` has `jQuery` as a dependancy. I don't quite follow. :smile:

themis_components/thSelect/readme.md
```diff
@@ -0,0 +1,156 @@
...
+  - A little too simple and nothing that we could build ourselves.
+  - untested
+- Chosen
+  - https://github.com/harvesthq/chosen
+  - Not too much activity on the repo lately.
+  - Heavy for what it does
+  - already in use in Clio
+- Select2
+  - https://github.com/select2/select2
+  - active. new version 4.0 around the corner with some nice features!
```
seanhealy: How far off is v4? What does it add? :)

themis_components/thSelect/readme.md
```diff
@@ -0,0 +1,156 @@
...
+  - lightweight. 72kb full minified
+- ui-select
+  - https://github.com/angular-ui/ui-select
+  - Angular wrapper for Select2
+  - **This could be a solution for us**. It can be rewritten with just what we need/want and slowly added to.
```
seanhealy: If we choose this option We'll still want to wrap it with a directive so we can keep the API consistent but this one could work.

themis_components/thSelect/readme.md
```diff
@@ -0,0 +1,156 @@
...
+
+
+
+
+### SIMPLE EXAMPLE
+---
+mockup example:
+````
+<th-select options="options" placeholder="pick one..." icon="car" label="Cars" ng-model='selectedCar'></th-select>
+```
```
seanhealy: Could you break the lines up in these examples? They are hard to read.

Also, when doing md here if you do
```md
  ```languagename
```

themis_components/thSelect/readme.md
```diff
@@ -0,0 +1,156 @@
...
+- Select
+  - http://github.hubspot.com/select/docs/welcome/
+  - A little too simple and nothing that we could build ourselves.
+  - untested
+- Chosen
+  - https://github.com/harvesthq/chosen
+  - Not too much activity on the repo lately.
+  - Heavy for what it does
+  - already in use in Clio
+- Select2
```
seanhealy: All the examples I've seen of Select 2 have that search bar in them. Do you know if that is required?

themis_components/thSelect/readme.md
```diff
@@ -0,0 +1,156 @@
...
+      url: 'https://api.clio.com/contacts'
+    ).done (data) ->
+      contact in data
+        arr.push contact.name
+            deferred.resolve(arr);
+    return deferred;
+  ```
+
+
+### SOME JS/JQ SELECTS AVAILABLE
```
seanhealy: Could you confirm the dependencies of each of these libs?

themis_components/thSelect/readme.md
seanhealy: Also, I'd like to see some use cases we think we will want based on what we currently have / need in the main app.

themis_components/thSelect/readme.md
JefCurtis: The search bar is *not* a requirement at all.

themis_components/thSelect/readme.md
JefCurtis: Looks like v4.0 just came out :sunglasses:
- Removing the need for hidden <input> elements
- Easier to style
- Advanced search matching

themis_components/thSelect/readme.md
JefCurtis: I was suggesting that we build our own wrapper for Select2 actually. I'll make that clear in the docs.

themis_components/thSelect/readme.md
JefCurtis: My bad. Just giving a quick idea. More pseudo code. Fixed!

themis_components/thSelect/readme.md
JefCurtis: If we opt to use Select2 with our select component we would requiring more external dependancies. Right now we don't require jQuery in order to use this library do we?

themis_components/thSelect/examples/2 - with ng-controller/coffee.coffee
```diff
@@ -0,0 +1,12 @@
+angular.module('thDemo', ['ThemisComponents'])
+  .controller "DemoController", ->
+    @data = {}
```
seanhealy: Any reason why you have spaced everything under `@data` here instead of just going `@options`, `@model`?

themis_components/thSelect/thSelect.directive.coffee
```diff
@@ -0,0 +1,71 @@
...
+
+      # when a new option is selected we want to capture the name
+      # and add it to our styled select replacement.
+      $element.on 'change', (event) =>
+        $scope.$apply =>
+          @selectedText = event.target.selectedOptions[0].text
+      return
+
+    link: (scope, element, attributes) ->
+      # grab the initially selected option and add it's name to our styled replacement select
```
seanhealy: Nitpicky: we should wrap our comments at about 80chars.

themis_components/thSelect/thSelect.directive.coffee
```diff
@@ -0,0 +1,71 @@
...
+    replace: true
+    bindToController: true
+    transclude: true
+    scope:
+      options: "="
+      ngModel: "="
+      name: "@"
+      disabled: "@"
+    controller: ($scope, $element) ->
+      @selectedText = @ngModel?.name ? "Choose..."
```
seanhealy: `Choose…` instead of `Choose...` the ellipsis character renders a bit differently / nicer.

themis_components/thSelect/thSelect.directive.coffee
```diff
@@ -0,0 +1,71 @@
...
+      $element.on 'change', (event) =>
+        $scope.$apply =>
+          @selectedText = event.target.selectedOptions[0].text
+      return
+
+    link: (scope, element, attributes) ->
+      # grab the initially selected option and add it's name to our styled replacement select
+      # this will only be applicable to if we are not passing in an array of options so we check for that first.
+      options = element.find("option")
+      if !attributes.options
```
seanhealy: `unless`

themis_components/thSelect/thSelect.directive.coffee
```diff
@@ -0,0 +1,71 @@
...
+        for option in options
```
seanhealy: for option in options when option.hasAttribute "selected"

seanhealy: What do we do if two elements have this?

themis_components/thSelect/examples/2 - with ng-controller/coffee.coffee
JefCurtis: In this case it was just a habit of not adding models directly to the scope but it isn't necessary so I can update that.

themis_components/thSelect/thSelect.directive.coffee
JefCurtis: Ha! that's funny that you say that because I literally went back and forth on that one. I'll switch.

themis_components/thSelect/thSelect.directive.spec.coffee
```diff
@@ -0,0 +1,115 @@
+context = describe
+describe "ThemisComponents: Directive: thSelect", ->
+  element = scope = compile = template = null
+
+  compileDir = (template, scope) ->
```
seanhealy: Do we not have a generic compile directive method?

themis_components/thSelect/thSelect.directive.coffee
JefCurtis: Nice.

themis_components/thSelect/thSelect.directive.spec.coffee
```diff
@@ -0,0 +1,115 @@
...
+    compiledElement = compile("<html ng-app>#{template}</html>")(scope)
+    scope.$digest()
+    {element: jQuery(compiledElement), scope: scope}
```
seanhealy: Do we have full jQuery here?
```coffeescript
return element: angular.element(compiledElement), scope: scope
```

themis_components/thSelect/thSelect.directive.spec.coffee
```diff
@@ -0,0 +1,115 @@
...
+
+  context "using an array of options", ->
+    beforeEach ->
+      module 'ThemisComponents'
+      inject ($rootScope, $compile) ->
+        scope = $rootScope.$new(); compile = $compile
+      scope.options = [{name: "Darth", value: "Sith Lord"}, {name: "Luke", value: "Jedi Knight"}]
```
seanhealy: hehe

themis_components/thSelect/thSelect.directive.spec.coffee
```diff
@@ -0,0 +1,115 @@
...
+      scope.choice = scope.options[0]
+      template = """<th-select name="faction" options="options" ng-model="choice"></th-select>"""
+      element = compileDir(template, scope).element
+
+    it "has the correct amount option elements", ->
```
seanhealy: correct *number of* option :tongue:

themis_components/thSelect/thSelect.directive.spec.coffee
```diff
@@ -0,0 +1,115 @@
...
+
+    it "value on select to equal select value", ->
+      expect(element.find("select")[0].value).toEqual "Sith Lord"
+
+    context "disabled", ->
+      beforeEach ->
+        disabled = """<th-select name="faction" options="options" disabled ng-model="choice"></th-select>"""
+        element = compileDir(disabled, scope).element
+
+      it "select is disabled", ->
```
seanhealy: Should we test that it doesn't work while disabled… Thinking about how to do this without an integration test…

themis_components/thSelect/thSelect.directive.coffee
JefCurtis: I guess it would always use the last `option:selected`. What should it do?

seanhealy: We should check what the HTML spec does in this case. Probably follow that convention. Or, throw an error…

themis_components/thSelect/thSelect.directive.spec.coffee
JefCurtis: Updated to spec helper to accepted additions to scope before compiling. I was getting a double injection error because I needed to inject `$scope` directly into my tests in order to create a new scope with `$rootScope.$new()` and then to add my required objects.

spec_helper/compileDirective.coffee
```diff
@@ -1,22 +1,22 @@
...
     compile = $compile
 
-  if options and options.promise
-    inject ($q) -> q = $q
-    scope.deferred = q.defer()
-    scope.promise = scope.deferred.promise
-
-  if options and options.action
-    scope.action = options.action
+  importScope() if scopeAdditions?
```
seanhealy: I know this isn't a change you made but as this is much more a function than an expression I think I would prefer an explicit return at the end. Thoughts?

Also, do you know when we added full jQuery to ThemisUI?

```coffee script
return element: jQuery(compiledElement), scope: scope
```

themis_components/thDropdown/thDropdown.directive.spec.coffee
```diff
@@ -63,8 +63,9 @@ describe "ThemisComponents: Directive: thDropdown", ->
 
   context "with ng-click", ->
     beforeEach ->
+      scope = {}; scope.action = -> alert 'response'
       template = """<th-dropdown name="Button" list="[{name:'action', ngClick: action}]"></th-dropdown>"""
-      element = compileDirective(template, {action: -> alert 'response'}).element
+      element = compileDirective(template, scope).element
```
seanhealy: Given your new setup here `scope` here looks as though it is being cloned into the `scope` being created by `compileDirective`. While this works as expected in this case it could cause oddness as your `scope` here is not pointing to the same in memory object as the scope used by compile. I like the though / plan here but, think we need to think about it a bit more…

themis_components/thSelect/thSelect.directive.coffee
```diff
@@ -0,0 +1,73 @@
...
+      $element.on 'change', (event) =>
+        $scope.$apply =>
+          @selectedText = event.target.selectedOptions[0].text
+      return
+
+    link: (scope, element, attributes) ->
+      # grab the initially selected option and add
+      # it's name to our styled replacement select
+      # this will only be applicable to if we are not
+      # passing in an array of options so we check for that first.
```
seanhealy: The spec on what to do in the event that more than one option on a non-multiselect select is set to selected is vague. I think we should perhaps throw a `console.warn` if this happens here. Since we can't be consistent with the spec we don't want people accidentally leaning on undefined behaviour.

spec_helper/compileDirective.coffee
JefCurtis: Ahh that's was added when Jordan created this helper. It makes writing tests more succinct I think and since it's only being used with the test it isn't a dependency on the library itself.

themis_components/thSelect/thSelect.directive.coffee
```diff
@@ -0,0 +1,79 @@
...
+      options = element.find "option"
+      unless attributes.options?
+        counter = 0
+        for option in options when option.hasAttribute "selected"
+          scope.select.selectedText = option.text
+          counter++
+        console.warn(
+          "#{counter} options are set on a non-multiple select (name: #{attributes.name}).
+           The last selected option will be used."
+        ) if counter > 1
```
seanhealy: The postfix if seems awkward here. How about:

```coffeescript
if counter > 1
  console.warn "#{counter} options are set on a non-multiple select (name: #{attributes.name}). The last selected option will be used."
```

spec_helper/compileDirective.coffee
seanhealy: :+1:

themis_components/thSelect/design.md
```diff
@@ -0,0 +1,193 @@
+# DESIGN DOCS
```
seanhealy: I really like that we did this. Thoughts on making it a more regular part of making a component?

JefCurtis: I think it's a good idea. Now that this simple select is finished do we pull out the irrelevant parts of the design doc?

PR #42, opened 2015-10-13
CLIO-26055: Added ability to add a class to th-popover

craigcarlyle:


themis_components/thPopover/thPopover.directive.coffee
```diff
@@ -28,6 +28,7 @@ angular.module('ThemisComponents')
       overlay = null
 
       templateURL = attributes.thPopover
+      $scope.class = attributes.popoverClass
```
seanhealy: Are you just wanting to pass a class into this directive for usage here? Want to understand the problem you're trying to solve.

PR #48, opened 2015-11-14
Clio-26384 Disclosure Component

andreisoare:


andreisoare:
Some issues still remaining on this PR, which I don't know how to fix yet:

 * I have 2 failing tests which should actually work
 * When you click on the toggle button, the window scroll goes up. I know what the cause is, but don't know how to fix it (it's related to ng-class).

themis_components/thDisclosure/index.scss
```diff
@@ -0,0 +1,16 @@
...
+}
+
+th-disclosure-toggle {
+  .fa-caret-right {
+    transform: none;
+    transition: transform 0.3s;
+  }
+
+  .fa-caret-right-rotated {
+    transform: rotate(90deg);
```
JefCurtis: No prefixes required here?

themis_components/thDisclosure/thDisclosureContent.directive.coffee
```diff
@@ -0,0 +1,41 @@
...
+  $(element).attr "style", previousCss ? ""
+  height
+
+angular.module 'ThemisComponents'
+  .directive 'thDisclosureContent', (DisclosureManager) ->
+    restrict: 'E'
+    transclude: true
+    scope:
+      name: '@'
+    template: require './thDisclosureContent.template.html'
```
JefCurtis: Why not place this inline? Not very big. Less mental overhead?
```html
<ng-transclude></ng-transclude>
```

themis_components/thDisclosure/thDisclosureContent.directive.coffee
```diff
@@ -0,0 +1,41 @@
...
+    bindToController: true
+    controllerAs: 'thDisclosureContent'
+    controller: ($element) ->
```
JefCurtis: Should we be doing the DOM manipulation inside `link`?

themis_components/thDisclosure/thDisclosureContent.directive.coffee
```diff
@@ -0,0 +1,41 @@
...
+  height
+
+angular.module 'ThemisComponents'
+  .directive 'thDisclosureContent', (DisclosureManager) ->
+    restrict: 'E'
+    transclude: true
+    scope:
+      name: '@'
+    template: require './thDisclosureContent.template.html'
+    bindToController: true
```
JefCurtis: Possible problem with scoping? I don't *think* the parent scope will be able to see this isolated scope?

themis_components/thDisclosure/thDisclosureManager.service.coffee
```diff
@@ -0,0 +1,12 @@
+angular.module 'ThemisComponents'
+  .factory 'DisclosureManager', ->
+    handlerMap = {}
+
+    onToggle = (name, handler) -> handlerMap[name] = handler
+
+    toggle = (name) -> (handlerMap[name] ? ->)()
+
+    return {
```
JefCurtis: Although the return is implicit here I still like to see it. Seems easier to read.

themis_components/thDisclosure/thDisclosureContent.directive.coffee
andreisoare: I think we should migrate all components to require templates rather than inline them. This helps with basic things like syntax highlighting in our templates and also to keep our component code clean. Big strings are ugly.

And I agree in this case there's no benefit, but for the purposes of consistency it's still worth doing in my opinion. There shouldn't be multiple ways of doing something unless there's a solid reason to do that.

@seanhealy thoughts?

themis_components/thDisclosure/index.scss
andreisoare: Yes, we need prefixes here - good catch.

@seanhealy do you know what the plans are for autoprefixer? Is it easy to add to ThemisUI? (like ... 1 point complexity?)

themis_components/thDisclosure/thDisclosureToggle.directive.coffee
```diff
@@ -0,0 +1,18 @@
...
+    template: require './thDisclosureToggle.template.html'
+    bindToController: true
+    controllerAs: 'thDisclosureToggle'
+    controller: ($element) ->
+      @expanded = false
+
+      @toggle = =>
+        @expanded = not @expanded
+        DisclosureManager.toggle @name
+        return
```
JefCurtis: Is the needed?

JefCurtis:
This component as it stands right now is leaving all the styling up to the implementer. Maybe the basic padding / layout / display should be styled while leaving the disclosure content itself a blank slate? Something similar to the way `th-popover` and `th-modal` work now. Thoughts?

themis_components/thDisclosure/thDisclosureContent.directive.coffee
andreisoare: Good question - I added a 3rd example to the documentation to show that it works.

themis_components/thDisclosure/thDisclosureToggle.directive.coffee
andreisoare: Yes :( the return value gets passed to the DisclosureToggle template and it throws random errors if it finds a value it doesn't like, so I'm just returning undefined.

andreisoare:
@seanhealy @PANCAKES what do you think about styling <th-disclosure-toggle>? Should we leave it generic, up to the user, or pre-style it with what we have right now for the Filter button in datatables? Is the disclosure paradigm used anywhere else besides filters?

themis_components/thDisclosure/thDisclosureManager.service.coffee
andreisoare: Agreed.

themis_components/thDisclosure/index.scss
andreisoare: Well, I just added it.

themis_components/thDisclosure/thDisclosureContent.directive.coffee
PANCAKES: I think I agree - as our components get more complex, we will want highlighting and a separation. +1 for consistency once we do want/need templates.

themis_components/thDisclosure/thDisclosureContent.directive.coffee
PANCAKES: Think I would prefer the ```$``` is removed here. You need to wrap it later in the function for jquery, so it is a bit confusing here IMO. Thoughts? Less chars!

themis_components/thDisclosure/thDisclosureToggle.directive.coffee
```diff
@@ -0,0 +1,18 @@
+angular.module 'ThemisComponents'
+  .directive 'thDisclosureToggle', (DisclosureManager) ->
+    restrict: 'E'
+    transclude: true
+    scope:
+      name: '@'
+    template: require './thDisclosureToggle.template.html'
+    bindToController: true
+    controllerAs: 'thDisclosureToggle'
+    controller: ($element) ->
```
PANCAKES: We don't need ```$element``` here if we don't use it

themis_components/thDisclosure/thDisclosureContent.directive.coffee
seanhealy: Off hand I think the $ is needed to have angular inject the element. But I also think that it will inject it as a jq element so the lower $() may be unneeded.

PANCAKES: Oh neat, #TIL. Did not know angular did that if you used the ```$``` here.

themis_components/index.scss
```diff
@@ -4,10 +4,11 @@
 // Oppa Component Style
 @import "thBindMarkdown/index";
 @import "thButton/index";
-@import "thDropdown/index";
```
PANCAKES: Nice. Alphabets

gulp/tasks/docsBrowserify.coffee
```diff
@@ -3,6 +3,8 @@ gulp = require 'gulp'
 browserify = require 'browserify'
 watchify = require 'watchify'
 source = require 'vinyl-source-stream'
+stringify = require 'stringify'
```
seanhealy: We'll have to set this up in Themis as well before we merge this guy but @JefCurtis is already doing this.

gulpfile.coffee
```diff
@@ -2,6 +2,7 @@ gulp = require 'gulp'
 path = require 'path'
 sass = require 'gulp-sass'
 rename = require 'gulp-rename'
+autoprefixer = require 'gulp-autoprefixer'
```
seanhealy: We'll need to add this in Themis as well.

package.json
```diff
@@ -17,11 +17,12 @@
     "npm": "^2.14.4"
   },
   "dependencies": {
-    "angular": "^1.3.15",
+    "angular": "1.4.7",
     "angular-animate": "^1.3.15",
```
seanhealy: The other `angular-*` libs should be updated too.

themis_components/thDisclosure/thDisclosureContent.directive.coffee
```diff
@@ -0,0 +1,41 @@
+$ = require 'jquery'
```
seanhealy: Why are you needing jQuery here?

themis_components/thDisclosure/thDisclosureContent.directive.coffee
seanhealy: I agree in general but I think for something this simple inline is nicer.

themis_components/thDisclosure/thDisclosureContent.directive.spec.coffee
```diff
@@ -0,0 +1,45 @@
...
+      element = compileDirective("""<th-disclosure-content name="unique-id">Content</th-disclosure-content>""").element
+
+    it "renders a <ng-transclude> component", ->
+      firstChild = getFirstChild element
+      expect(firstChild.is('ng-transclude')).toBe true
+
+    it "has height 0", ->
+      expect(element.css("height")).toEqual "0px"
+
+    it "transcludes its children", ->
```
seanhealy: We're testing angular's internals here. Is this really necessary do you think?

themis_components/thDisclosure/thDisclosureContent.directive.coffee
andreisoare: For `$($element).animate`. jQuery doesn't exist globally in the app, so `$element` uses Angular's jqLite.

themis_components/thDisclosure/thDisclosureToggle.directive.coffee
seanhealy: Don't quite follow. What error are you seeing with out this `return`?

themis_components/thDisclosure/thDisclosureManager.service.coffee
```diff
@@ -0,0 +1,12 @@
+angular.module 'ThemisComponents'
+  .factory 'DisclosureManager', ->
+    handlerMap = {}
+
+    onToggle = (name, handler) -> handlerMap[name] = handler
```
seanhealy: Wondering if it might be good to throw and error if we try to add a non-unique handler here…

themis_components/thDisclosure/thDisclosureContent.directive.coffee
andreisoare: It is easier when writing it, but harder when reading code. And we read way more code than we write.

I remember Kyle mentioning this during the summit - ideally the developer should know exactly where to find code for a module (for example: the template for a specific component, or the test for a specific controller in the case of Rails).

If we have 2 ways, then the developer will always need to search for templates in 2 places. It's a small consistency detail.

I know that thDisclosure is actually creating the inconsistency in this case, so if you want I can create another PR that moves all the other templates to external files.

What do you think?

themis_components/thDisclosure/thDisclosureManager.service.coffee
andreisoare: I initially wanted to do that but then I thought of the following scenario:

* User visits route A that renders a component that uses thDisclosure with a unique id.
* User visits route B that renders another component instead
* User comes back to route A, so the component associated with route A is rebuilt, thus registering a new handler for the same unique id. We don't want an error in this case.

Do you find anything wrong with this scenario?

themis_components/thDisclosure/thDisclosureContent.directive.spec.coffee
andreisoare: Agreed, this is useless.

themis_components/thDisclosure/thDisclosureToggle.directive.coffee
andreisoare: This one: https://docs.angularjs.org/error/$parse/isecdom?p0=thDisclosureToggle.toggle(

themis_components/thDisclosure/thDisclosureContent.directive.coffee
seanhealy: It should still be creating a jq lite element. However your point about animate stands. We could link in jQuery earlier so that we get proper jQuery elements here. Thoughts?

themis_components/thDisclosure/thDisclosureManager.service.coffee
seanhealy: I was pondering that same problem when I posted my previous comment but my brain was fried and I thought I had double thought myself. I would still like to have some protection against this thinking about how we could accomplish this…

themis_components/thDisclosure/thDisclosureManager.service.coffee
seanhealy: Also agree.

themis_components/thDisclosure/thDisclosureToggle.directive.coffee
seanhealy: Hmmm… weird… tis all good though. :)

themis_components/thDisclosure/thDisclosureContent.directive.coffee
seanhealy: I'm torn. For something this simple I'm not a fan of having to open a separate file. But I do see your point…

themis_components/thDisclosure/thDisclosureContent.directive.coffee
seanhealy: Commented on this above.

themis_components/thDisclosure/thDisclosureManager.service.coffee
andreisoare: Here are some half-baked ideas that I have, maybe you can get them to the finish line:

1. Force the developer to use unique ids and throw errors if he doesn't. Then it's on him to generate unique ids when components are rebuilt (maybe the ids would depend on some component-specific data).

2. Reset the handlerMap[name] whenever th-disclosure-toggle initializes. But this won't work if toggle appears after content on the page. It also won't work if only content gets re-built and toggle lives forever.

themis_components/thDisclosure/thDisclosureContent.directive.coffee
andreisoare: I know we talked about using jQuery as a Common.js module and not a global, which is why I avoided putting it in the global scope.

If we want to keep the jQuery footprint minimal and at some point remove it completely (ngAnimate [is getting a revamp in Angular 2](https://www.youtube.com/watch?v=izp5N2eeogg), and I'm pretty sure it will help us get rid of $.animate), then I think we shouldn't embed it in Angular, because we will make it convenient for developers to abuse it.

What do you think?

themis_components/thDisclosure/thDisclosureContent.directive.coffee
andreisoare: The other thing is - if we keep 2 patterns for doing this then there will always be cases where both will work (say, 2 lines of template code) and we'll just have a mix of them. Also what happens when an inline template grows slightly? Do you have to pull it out into a separate file?

It just creates mental overhead.

seanhealy: You've convinced me. :laughing:

themis_components/thDisclosure/thDisclosureContent.directive.coffee
seanhealy: Yea, I'll agree. We'll have to check how this plays in themis but I reckon it should be fine.

seanhealy:
Lets chat re styling at the same time as the above point.

PR #49, opened 2015-11-16
Clio 26383 header component

PANCAKES:


package.json
```diff
@@ -22,6 +22,7 @@
     "angular-route": "~1.3.0",
     "angular-sanitize": "^1.3.15",
     "express-ws": "^0.2.6",
+    "flexbox": "0.0.3",
```
seanhealy: What is this?

themis_components/thContentHeader/thContentHeader.directive.coffee
```diff
@@ -0,0 +1,11 @@
+angular.module('ThemisComponents')
+  .directive 'thContentHeader', ->
+    restrict: "AE"
+    scope:
+      title: "@"
+    transclude: true
+    template: """
+      <h1 class="th-header-title th-header-flex-item">{{ title }}</h1>
```
seanhealy: Just so we stay in the habit it would be good to use the `controllerAs` syntax here.

PR #52, opened 2015-11-19
clio-26900 DataTable component design proposal

andreisoare:


themis_components/thTable/readme.md
```diff
@@ -0,0 +1,70 @@
+# DataTable — `thTable` / `thTableCell` / `thTableRowExtension`
+
+### Description
+
+I think tables and lists should be separate components. Why? Because if we make
+them a single generic component, then we will have to make compromises in their
+usage (it gets more complicated that it should be).
```
seanhealy: I think this is rational. In this case I think we should not make a list component at this time.

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,70 @@
...
+
+  <th-table-cell [headerTitle="Title for the entire column"] [sortable="fieldName"]>
+    [...]
+  </th-table-cell>
+
+  <th-table-row-extension position="below(default)|above">
+    You can add an optional row extension that is displayed below or above each
+    set of cells. This is where you would add action buttons for the current
+    {{item}} in the row.
+  </th-table-row-extension>
```
seanhealy: This part feels a bit funny to me. That doesn't mean it's bad I just want to think it over for a little bit here.

andreisoare: Another alternative would be to not have a special table-row-extension and just put other content before and/or after the th-table-cell enumeration and the component would know to extract those and render them before and/or after the cell line. But this added flexibility might be more confusing.

andreisoare:
Picture this table when thinking about the design for datatables. This is the most complex I could find in Clio.

screenshot 2015-11-18 16 52 50

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,70 @@
...
+
+I think tables and lists should be separate components. Why? Because if we make
+them a single generic component, then we will have to make compromises in their
+usage (it gets more complicated that it should be).
+
+
+### List component w/ pagination
+
+```html
+<th-list objects="arrayOfObjects"
```
doliveirakn: Is the intention to have a th-list inside of a th-table?

andreisoare: No. We initially talked about making th-list and th-table a single component that's customizable enough to act as both, but then I tried to do that and wasn't able to do it without compromising API simplicity.

I just added the design for `th-list` here for an idea of how it will look like, but they are completely separate components. And we don't need `th-list` right away, so will not even implement it.

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,70 @@
...
+         [objectReference="item"]
+         [pageNumber="1"]
+         [totalPages="10"]
+         [changePage="(page) ->"]>
```
doliveirakn: Seems odd to have an embedded function in the HTML. Though maybe this is just me not understanding the format of the documentation. Just for clarification, which would it look like
1:
```
<th-list changePage="function(page){ ... }"></th-list>
```
2:
 ```
<th-list changePage="functionReference"></th-list>
```

File: themis_components/thTable/readme.md [hunk 5]
```diff
@@ -0,0 +1,70 @@
...
+usage (it gets more complicated that it should be).
+
+
+### List component w/ pagination
+
+```html
+<th-list objects="arrayOfObjects"
+         [objectReference="item"]
+         [pageNumber="1"]
+         [totalPages="10"]
```
doliveirakn: Would it be better to specify page size rather than total pages?
We may need to make requests like `limit=X`, and given the page size and the `arrayOfObjects.length` we should be able to calculate the total number of pages

themis_components/thTable/readme.md
andreisoare: The 2nd one. I was just describing the function signatures, sorry for the confusion.

themis_components/thTable/readme.md
doliveirakn: for QA purposes, it would be useful to be able to specify an html ID prefix/identifier for each list item. So we could end up with something like

`<tr id=matters-X>...</tr>` This would allow QA to easily hunt down rows for any list

themis_components/thTable/readme.md
andreisoare: `arrayOfObjects.length` will not be the total number of objects, it will just be the total number of objects that the client has loaded so far.

We can't load all objects in the client with 1 network request, can we?

As for page size and the arrayOfObjects, the way I imagine it is that the table component doesn't really care about page size, it just receives the elements for the current page to display via arrayOfObjects.

themis_components/thTable/readme.md
doliveirakn: I agree that the the functionality is useful. 

The thing that feels funny to me is that it is mixed up with the `th-table-cell`. Theoretically, in the HTML, i could define 2 table cells, then a table-row-extension, then 2 more table cells. This feel conceptually weird. Not sure the best way to get around that.

themis_components/thTable/readme.md
doliveirakn: So we need to be able to sort a list? if so, we should probably also include the `onSort` attribute from the table

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,70 @@
...
+<th-table objects="arrayOfObjects"
+          [objectReference="item"]
+
+          [pageNumber="1"]
+          [totalPages="10"]
+          [changePage="(page) ->"]
+
+          [onSort="(field, order) ->"]>
+
+  <th-table-cell [headerTitle="Title for the entire column"] [sortable="fieldName"]>
```
doliveirakn: I know in the past, we've had to add classes specifically to certain `<th>` elements in the datatables. This looks to abstract the creation of the `<th>` elements (which is fine) but we might want a want to put classes/ids/attributes onto the those elements

themis_components/thTable/readme.md
doliveirakn: Good point re: total number of objects. I was getting ahead of myself and forgot that `arrayOfObjects` was already a subset of information. 

I do think though that rather than `totalPages` we'd want a total items/items per page. Otherwise I think we'd be duplicating logic in whatever is initializing (ie. we'd need to duplicate the calculate total number of pages functionality)

themis_components/thTable/readme.md
andreisoare: In the scenario you presented the component could throw an error.

themis_components/thTable/readme.md
seanhealy: I have some thoughts here. I'll write them up once I'm done with scansnap stuff from this morning.

andreisoare: I haven't seen a table in Clio that would need this. Can you think of an example?

If we indeed need to be able to customize headers, then the design of this component must change.

themis_components/thTable/readme.md
doliveirakn: Cool, if there are mechanisms that could be in place to prevent it, I would be happy.

themis_components/thTable/readme.md
doliveirakn: I think it is less with the columns now that I"m thinking about it, and more the `<td>` element (eg. add a class so that the text doesn't wrap, or does wrap)

fimmtiu: Things I know we've manually specified on table headers before:
* Disable sorting on that header.
* `white-space: nowrap` to prevent line breaks in the headers.
* Left or right justification of the header.

themis_components/thTable/readme.md
fimmtiu: Agreed with Kyle here. Total items and items per page feel more versatile; you can trivially derive the number of pages from `ceil(total / per_page)`, but given the total number of pages you can't derive the number of items.

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,70 @@
...
+
+```html
+<th-table objects="arrayOfObjects"
+          [objectReference="item"]
+
+          [pageNumber="1"]
+          [totalPages="10"]
+          [changePage="(page) ->"]
+
+          [onSort="(field, order) ->"]>
```
fimmtiu: How about a way to specify the default sort order (ASC vs. DESC)? We do that in some places, IIRC -- look up the `sorting_desc` CSS class.

themis_components/thTable/readme.md
fimmtiu: How about a way to specify the default sort order (ASC vs. DESC)? We do that in some places, IIRC -- look up the sorting_desc CSS class.

andreisoare: It seems the only use case for customizing header cells is for alignment, so I added a `header-align` attribute.

Also added a way to specify default sort order.

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,60 @@
...
+    the current object in the row.
+  </th-table-cell>
+
+  ... <!-- Define all <th-table-cell>s first. -->
+
+  <th-table-cell ...> ... </th-table-cell>
+
+  <!-- Then define an optional row extension for things like action buttons. -->
+
+  <th-table-row-extension>
```
doliveirakn: I think we still need an option to say if this is before, or after the table cells. (either that or the description is off)

themis_components/thTable/readme.md
seanhealy: Agree with having an identifier but I think it would be better for us to standardize a custom attribute than re-purpose `id`. `id` has other implications that don't necessarily fit this case but, it is a very good idea. Perhaps we could do something like `data-identifier`.

themis_components/thTable/readme.md
seanhealy: We've made a conscious effort in our components so far to avoid exposing classes that people can use to modify them on a one-off basis.

One-offs like this are partly to blame for the unpredictable state of our current UI. As soon as components in use are behaving in or styled in ways that they were't designed in then we can to longer rely on them behaving predictably when changing the component itself nor can users predict its behaviour.

There will be times when we need additionally functionality on a component but this is almost always best done as an enhancement to the component itself.

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,67 @@
...
+          [total-objects="233"]
+          [change-page="(page) ->"]
+
+          [on-sort="(field, order) ->"]>
+
+  <!-- First define all <th-table-cell>s that you want to display on
+       each table row.
+    -->
+
+  <th-table-cell [header-title="title for the entire column"]
```
seanhealy: How are we able to control the sizing of the table cells? Do we want to provide those using this component to specify that? I would think so…

andreisoare: Yes, it can be done here. Either as a class on the element, or inside the element. We can probably make this decision better when we integrate the datatable in the admin panel.

PR #53, opened 2015-11-19
Clio 21025 th loader

JefCurtis:
- Updated spec helper to now allow multiple injections while writing tests. 
- Small changed but took some time to come to a fix. 
 -Update for thLoader readme. 
- CR comments addressed.
- The styling for the loader uses `em`s as this allows for flexibility in it's size but only needing to change the font size.


spec_helper/compileDirective.coffee
```diff
@@ -12,7 +12,7 @@ window.compileDirective = (template, scopeAdditions) ->
       scope[key] = value
 
   # Inject
-  inject ($rootScope, $compile) ->
+  inject ($rootScope, $compile, $q, $timeout) ->
```
seanhealy: Do we still need `$q` and `$timeout` here?

themis_components/thLoader/index.scss
```diff
@@ -0,0 +1,67 @@
...
+.loading-text {
+  color: $dark-grey-3;
+  @include gibson-semibold;
+}
+
+@keyframes spin {
+  to { transform: rotate(1turn); }
+}
+
+.progress {
```
seanhealy: we should nest all these rules under `.th-loader` so they don't get applied in other unexpected places.

JefCurtis: Right good call. On it.

themis_components/thLoader/index.scss
```diff
@@ -0,0 +1,67 @@
...
+
+  .progress > div:before {
+    transform: rotate(-90deg);
+    background: #999;
+  }
+
+  .progress > div:after {
+    transform: rotate(-135deg);
+    background: #bbb;
+}
```
seanhealy: This `}` should be indented. Nit picky. :p

PR #54, opened 2015-11-25
Clio 26902 - Implement thTable

andreisoare:


themis_components/thTable/examples/3 - Custom Sort And Pagination/coffee.coffee
```diff
@@ -0,0 +1,80 @@
...
+  .controller "DemoController", (SimpleTableDelegate, TableHeader) ->
+    data = fixtures 102
+
+    pageSize = 5
+
+    headers = [
+      new TableHeader
+        name: 'First Name'
+        sortField: 'firstName'
+        sortEnabled: 'ascending'
```
seanhealy: `sortEnabled` here looks to me like it's controlling whether or not I can sort. I get the impression though that it's controlling the initial sorting state…?

themis_components/thTable/examples/3 - Custom Sort And Pagination/coffee.coffee
```diff
@@ -0,0 +1,80 @@
...
+      new TableHeader
+        name: 'Last Name'
+        sortField: 'lastName'
+    ]
+
+    sortByHeader = (header) ->
+      data.sort (a, b) ->
+        result = a[header.sortField].localeCompare b[header.sortField]
+        result = -result if header.sortEnabled is "descending"
+        result
```
seanhealy: I think this is a bit big for an expression. `return result`

themis_components/thTable/examples/3 - Custom Sort And Pagination/coffee.coffee
```diff
@@ -0,0 +1,80 @@
...
+      header = getCurrentSortHeader()
+      start = (page - 1) * pageSize
+      end = start + pageSize
+      updateData {
+        data: getDataPage sortByHeader(header), page
+      }
+
+    initialData = getDataPage sortByHeader(getCurrentSortHeader()), 1
+
+    @tableDelegate = new SimpleTableDelegate {
```
seanhealy: I think it would be clearer to declare these fields inside the creation of the delegate rather than above and then linked in. It's just hard to tell what is relevant to the table and what isn't. This will get worse in a real controller where there are mixed pieces.

themis_components/thTable/thSimpleTableDelegate.service.coffee
```diff
@@ -0,0 +1,173 @@
...
+      headers
+      onSort
+      updateData: updateDataFromSort
+    }
+
+    sortData = (header) ->
+      updatePagination {currentPage: 1}
+      sort data, header
+
+    post = (rows) ->
```
seanhealy: I understand where this post name is coming from but I think we can be better about it's naming / description. It requires a decent understanding of angular and directives for this name to make sense.

themis_components/thTable/examples/3 - Custom Sort And Pagination/coffee.coffee
andreisoare: Agreed, poor choice of name.

Right now:
 * a column is considered sortable is `sortField` is found.
 * a column is actively sorted by if `sortEnabled` is set to either ascending or descending.

Should I make `sortEnabled` true/false and add another field called `sortDirection`?

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
```diff
@@ -0,0 +1,58 @@
+fixtures = ->
+  celebrities = [
+    {id: 1, name: "Jose Valim",  twitter: "josevalim"}
+    {id: 2, name: "Dan Abramov", twitter: "dan_abramov"}
+    {id: 3, name: "Todd Motto",  twitter: "toddmotto"}
+    {id: 4, name: "Yehuda Katz", twitter: "wycats"}
```
seanhealy: But Yehuda Katz is mean! https://news.ycombinator.com/item?id=3832575 :tongue:

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
```diff
@@ -0,0 +1,58 @@
...
+          sortDirection: 'ascending'
+
+        TableHeader
+          name: 'Twitter'
+          sortField: 'twitter'
+      ]
+
+      fetchData: (page, pageSize, sortHeader, updateData) ->
+        sortedData = sort data, sortHeader
+        updateData undefined, sortedData, sortedData.length
```
seanhealy: Thoughts on passing named params to `updateData`? You seem to be ignoring the first parameter a lot.

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
```diff
@@ -0,0 +1,58 @@
...
+      headers: [
+        TableHeader
+          name: 'Id'
+          sortField: 'id'
+
+        TableHeader
+          name: 'Name'
+          sortField: 'name'
+          sortActive: true
+          sortDirection: 'ascending'
```
seanhealy: I was talking to Jef a while back about defining constants for some of the keywords we have like this. I'm not totally sure how I feel about it and we do it a lot so we don't necessarily need to start doing it here just thought I'd see what you think.

Sort of along these lines: https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIActivityIndicatorView_Class/index.html#//apple_ref/c/tdef/UIActivityIndicatorViewStyle

themis_components/thTable/examples/1 - Actions Row Sort/html.html
```diff
@@ -0,0 +1,27 @@
...
+          {{item.name}}
+        </th-table-cell>
+
+        <th-table-cell>
+          <a target="_blank" href="{{item.twitterProfile}}">{{item.twitter}}</a>
+        </th-table-cell>
+      </th-table-row>
+
+      <th-table-row type="actions" start-column="3">
+        <a target="_blank" href="{{item.twitterFollowing}}">Following</a>
```
seanhealy: Nit-picky but this is slightly more correct: https://docs.angularjs.org/api/ng/directive/ngHref

themis_components/thTable/thSimpleTableDelegate.service.coffee
```diff
@@ -0,0 +1,176 @@
...
+  interpolateStart = $interpolate.startSymbol()
+  interpolateEnd = $interpolate.endSymbol()
+
+  SimpleTableDelegate = (options) ->
+    delegate = TableDelegate options
+
+    {
+      generatePagination
+      headers
+    } = delegate
```
seanhealy: Every time I read this I get confused as to where `delegate` is coming from. It's weird I know as it's defined immediately above. At first I was thinking about suggesting:

```coffeescript
{
  generatePagination
  headers
} = TableDelegate options
```

But, it seems that you need the delegate reference for later… Likely no action is required I'm just rambling a bit… :p

themis_components/thTable/thTableDelegate.service.coffee
```diff
@@ -0,0 +1,85 @@
+angular.module 'ThemisComponents'
+.factory 'TableDelegate', (TablePagination) ->
+  SimpleTableDelegate = (options) ->
+    {
+      headers
+      currentPage
+      pageSize
+      fetchData
+    } = options
```
seanhealy: Should we set any defaults here?

Setting `headers` to `[]` if it doesn't exist would make the code lower a bit more readable I think…

```coffeescript
{
  headers = []
  currentPage
  pageSize
  fetchData
} = options
```

themis_components/thTable/thTableDelegate.service.coffee
```diff
@@ -0,0 +1,85 @@
...
+
+    data = []
+    loading = false
+    error = false
+
+    currentSortHeader = (headers ? []).find (header) -> header.isSortActive()
```
seanhealy: See above.

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
```diff
@@ -0,0 +1,58 @@
...
+
+        TableHeader
+          name: 'Twitter'
+          sortField: 'twitter'
+      ]
+
+      fetchData: (page, pageSize, sortHeader, updateData) ->
```
seanhealy: Really liking this guy.

I'm understanding that `updateData` is effectively a callback correct?

themis_components/thTable/thTableDelegate.service.coffee
```diff
@@ -0,0 +1,85 @@
...
+      getCurrentPage
+      getPageSize
+    } = TablePagination {
+      currentPage
+      pageSize
+      onChangePage: triggerFetchData
+    }
+
+    sortData = (header) ->
+      return if not header.sortField
```
seanhealy:
```coffeescript
return unless header.sortField?
```

themis_components/thTable/thTableDelegate.service.coffee
```diff
@@ -0,0 +1,85 @@
...
+
+    getData = -> data
+    getError = -> error
+    isLoading = -> loading
+    isEmpty = -> data.length is 0 and !isLoading() and !getError()
+
+    triggerFetchData()
+
+    return Object.freeze {
+      getData
```
seanhealy: For the functions that aren't used internally like `getData` thoughts on defining them inline here?

I guess maybe we might need some logical readability separation between the public signature of this class and it's private methods…

themis_components/thTable/thTableHeader.service.coffee
```diff
@@ -0,0 +1,67 @@
+angular.module 'ThemisComponents'
+  .factory 'TableHeader', -> TableHeader
+
+TableHeader = (options) ->
+  {
+    name
+    sortField
+    sortActive
+    sortDirection
+    align
```
seanhealy: `align = "left"`

themis_components/thTable/thTableHeader.service.coffee
```diff
@@ -0,0 +1,67 @@
...
+  } = options
+
+  align = align ? "left"
```
seanhealy: Don't need this.

themis_components/thTable/thTablePagination.service.coffee
```diff
@@ -0,0 +1,117 @@
...
+
+  TablePagination = (options) ->
+    {
+      currentPage
+      pageSize
+      onChangePage
+    } = options
+
+    totalItems = 0
+    dotdotdot = '...'
```
seanhealy: `ellipsis = '…'`

themis_components/thTable/thTablePagination.service.coffee
```diff
@@ -0,0 +1,117 @@
...
+    maxConsecutivePages = 5
+
+    totalPages = -> Math.ceil totalItems / pageSize
+
+    pages = ->
+      return [] unless totalItems > 0
+
+      lastPage = totalPages()
+      if lastPage <= maxConsecutivePages + 4
```
seanhealy: Should we set `4` here to be less of a magic number.

themis_components/thTable/thTablePagination.service.coffee
```diff
@@ -0,0 +1,117 @@
...
+        return [1 .. lastPage]
+
+      if maxConsecutivePages % 2 is 0
+        start = currentPage - maxConsecutivePages / 2 + 1
+        end = currentPage + maxConsecutivePages / 2
+      else
+        start = currentPage - Math.floor maxConsecutivePages / 2
+        end = currentPage + Math.floor maxConsecutivePages / 2
+
+      if start < 3
```
seanhealy: Same here re `3`.

themis_components/thTable/examples/3 - Async Sort And Pagination/coffee.coffee
```diff
@@ -0,0 +1,59 @@
...
+
+        TableHeader
+          name: 'Last Name'
+          sortField: 'lastName'
+      ]
+
+      currentPage: 1
+      pageSize: 5
+
+      fetchData: (page, pageSize, sortHeader, updateData) ->
```
seanhealy: Pondering parameters here. `updateData` is the only field that is actually required. The others all provide meta data so you can properly provide updateData with what it needs to function. I really like the thought here; just wondering how this will scale. Perhaps we could do a signature along the lines of:

```coffeescript
fetchData: (update, meta) ->
  {
    page
    pageSize
    sortHeader
  } = meta
```

This way you could *request* the meta data that you want to make the function relevant for your usage…

themis_components/thTable/examples/5 - No Data/html.html
```diff
@@ -0,0 +1,20 @@
...
+          {{item.id}}
+        </th-table-cell>
+
+        <th-table-cell>
+          {{item.name}}
+        </th-table-cell>
+      </th-table-row>
+
+      <th-table-row type="no-data">
+        Sir, I can't find any data!
```
seanhealy: hehe

themis_components/thTable/index.scss
```diff
@@ -0,0 +1,151 @@
...
+  position: relative;
+
+  .th-table-loading {
+    opacity: 0.5;
+  }
+
+  &:after {
+    content: "";
+    display: table;
+    clear: both;
```
seanhealy: Why do we need a clear fix here?

themis_components/thTable/index.scss
```diff
@@ -0,0 +1,151 @@
...
+      }
+    }
+
+    tr.th-table-hover-row {
+      background: #fafdff;
+    }
+  }
+
+  .th-table-pagination {
+    float: right;
```
seanhealy: Is this just here to push the page controls to the right?

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,242 @@
...
+      You can use {{item}} or whatever is configured by object-reference to access
+      the current object in the row.
+    </th-table-cell>
+    <th-table-cell>...</th-table-cell>
+    <th-table-cell>...</th-table-cell>
+  </th-table-row>
+</th-table>
+```
+
+## Old design
```
seanhealy: Before we merge this we should probably pull this out as it will just confuse people.

themis_components/thTable/thSimpleTableDelegate.service.coffee
```diff
@@ -0,0 +1,176 @@
...
+          </div>
+        </div>
+      """
+
+    generateHeaders = ->
+      return "" unless (headers or []).length > 0
+      template = """
+        <thead>
+          <tr>
+            <th ng-repeat="header in thTable.delegate.headers track by $index"
```
seanhealy: Do you foresee any case where the header order would change? If so we might want to be wary of tracking by `$index`… Donno…

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
andreisoare: I agree, he is mean sometimes :( but  he still builds great things!

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
andreisoare: Good idea!

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
andreisoare: Yeah I hate having to manually type string constants - it's very error prone. I do throw an exception if the strings aren't recognized.

The alternative I was proposing here was to replace it with a `sortAscending` flag (true / false), but we talked about that and decided to avoid it.

I am in favour of using string constants, but the issue I have is that I'm not sure of an elegant way to import them, because we're using Angular modules instead of requires. In ThemisUI it's not a problem - I just add a `constants.coffee` file in thTable and require it, but in Themis it would be hard to require.

themis_components/thTable/thSimpleTableDelegate.service.coffee
```diff
@@ -0,0 +1,176 @@
...
+        </thead>
+      """
+
+    generateBody = (rows) ->
+      numColumns = childrenArray(rows['cells']).length
+      cellsRow = generateCellsRow rows['cells'], rows['actions']?
+      actionsRow = generateActionsRow rows['actions'], numColumns
+      noDataRow = generateNoDataRow rows['no-data'], numColumns
+      errorRow = generateErrorRow numColumns
+      template = """
```
seanhealy: One thought on this syntax. While `this` is annoying in normal JS classes it does nicely serve to differentiate between variables scoped to a function and those scoped to the instance. Pondering a convention we could use around this to help increase readability.

themis_components/thTable/thSimpleTableDelegate.service.coffee
```diff
@@ -0,0 +1,176 @@
...
+        </tbody>
+      """
+
+    generateCellsRow = (cellsRow, hasActionsRow) ->
+      ngRepeat = if hasActionsRow then "ng-repeat-start" else "ng-repeat"
+      objectReference = getObjectReference cellsRow
+      cells = childrenArray(cellsRow)
+                .map (cell) -> generateCell cell
+                .join ''
+      template = """
```
seanhealy: It looks to me like you are just returning this string. Is there a reason you are also storing it to `template`?

themis_components/thTable/thSimpleTableDelegate.service.coffee
```diff
@@ -0,0 +1,176 @@
...
+                We had trouble loading your content. <br>
+                <a href ng-click="thTable.delegate.triggerFetchData()">Try again</a>.
+              </span>
+            </div>
+          </td>
+        </tr>
+      """
+
+    checkValidRows = (rows) ->
+      if not rows["cells"]?
```
seanhealy: `unless`

themis_components/thTable/thSimpleTableDelegate.service.coffee
```diff
@@ -0,0 +1,176 @@
...
+                        "for the actions and cells rows."
+
+      if actionsRow?
+        startColumn = parseInt(actionsRow.getAttribute('start-column')) || 1
+        numColumns = childrenArray(cellsRow).length
+        if startColumn > numColumns
+          throw new Error "start-column cannot be bigger " + \
+                          "than the number of cells"
+
+    childrenArray = (node) ->
```
seanhealy: I'm understanding that you are using this to get and convert an `HTMLCollection` into an `Array`?

themis_components/thTable/thTableDelegate.service.coffee
andreisoare: Neat, I wished there was a way to do this in an elegant way. Didn't know about this syntax.

themis_components/thTable/thSimpleTableDelegate.service.coffee
seanhealy: After thinking a bit more about this this could just be a coffeescript syntax issue as in pure js we would see `var` or `let`.

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
andreisoare: Yes. Works great with async.

themis_components/thTable/thTablePagination.service.coffee
andreisoare: Haha, that works better. I was inspired by [dotdotdot.js](http://dotdotdot.frebsite.nl/) :smile:

themis_components/thTable/thTablePagination.service.coffee
andreisoare: 4 comes from the following thought process: we always show the first page and the last page, and if the current page is far from both the first page and the last one, then we also show 2 ellipses. So if we have a total of 9 pages it wold be stupid to show: 1 ... 3 4 5 6 7 ... 9

How do I summarize the above sentence into one word? :crying_cat_face: 

Do you want me to add the above as a comment in the code?

themis_components/thTable/index.scss
andreisoare: Pagination is floated, so it isn't counted in the total height of th-table. I need th-table to know about its correct height because I'm adding an overlay to block the table controls while `isLoading() is true`.

themis_components/thTable/index.scss
andreisoare: Yes, to show pagination on the right side below the table.

themis_components/thTable/readme.md
andreisoare: I will completely rewrite the documentation, as a lot has changed even from design v2.

themis_components/thTable/thSimpleTableDelegate.service.coffee
andreisoare: The problem here is that headers, like all the other objects, are frozen. ng-repeat was trying to add a field to them, which is why I explicitly set `track by`.

I don't know why header order would ever change, I don't expect that to happen. If you think we might need this, I can figure something out to fix this issue. Let me know.

themis_components/thTable/thSimpleTableDelegate.service.coffee
andreisoare: No real reason, just for readability. I can replace them with return statements if you want.

themis_components/thTable/thSimpleTableDelegate.service.coffee
andreisoare: Yes, I needed that so I could call map on the array. So stupid that they're not arrays by default.

themis_components/thTable/thSimpleTableDelegate.service.coffee
andreisoare: It is confusing, I agree, you don't know if a variable is local or global.

But like you said, it's a coffeescript fail - this problem exists everywhere you define functions. We would have to create a convention to always use `this` for local variables.

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
andreisoare: Think I should rename it to `next`, `done`, `cbk`, or something else?

themis_components/thTable/examples/3 - Async Sort And Pagination/coffee.coffee
andreisoare: Good idea.

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
seanhealy: Sorry to mis lead my question wasn't with regard to its name. I was thinking about something related and wanted to confirm what I thought was happening here. :)

themis_components/thTable/examples/3 - Async Sort And Pagination/coffee.coffee
seanhealy: :)

Context might be a better word then meta though…

themis_components/thTable/index.scss
seanhealy: Why don't we use flex-box for this instead then. It's less hacky than abusing float like this.

themis_components/thTable/readme.md
seanhealy: :+1:

themis_components/thTable/thSimpleTableDelegate.service.coffee
seanhealy: Lets not worry about this for now. I was thinking way in the future where we may let users reorder headers. :tongue:

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
andreisoare: My question remains - do you think I should rename it, or is `updateData` a good name?

themis_components/thTable/examples/3 - Async Sort And Pagination/coffee.coffee
andreisoare: I was thinking `options`? It's consistent with all the other functions I defined.

themis_components/thTable/thSimpleTableDelegate.service.coffee
andreisoare: Haha, sounds like a feature in year 2020 :rocket:

themis_components/thTable/thSimpleTableDelegate.service.coffee
```diff
@@ -0,0 +1,173 @@
...
+      arr = []
+      arr.push(child) for child in node.children
```
andreisoare: @seanhealy is this the proper way to get a raw node's children? Will it work in all browsers? I never really knew if the DOM API is the same in all browsers and what's safe to use.

seanhealy: I would just do:

```coffeescript
childrenArray = (node) ->
  return [].slice.call(node.children)
```

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
seanhealy: :cry:

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
seanhealy: I was pondering more about the possibility of doing a promise return or something but… I think that that would be too much overhead in this case.

The name was not a concern. :)

themis_components/thTable/thTablePagination.service.coffee
seanhealy: Looks good.

themis_components/thTable/examples/3 - Async Sort And Pagination/coffee.coffee
seanhealy: I was thinking `context` only because the table is making context about itself available to make decisions around. Don't feel strongly about this though.

themis_components/thTable/thSimpleTableDelegate.service.coffee
seanhealy: The distant future. The year 2000!

themis_components/thTable/thSimpleTableDelegate.service.coffee
seanhealy: :smile:

themis_components/thTable/thSimpleTableDelegate.service.coffee
seanhealy: Yea, should be good in all the browsers we care about.

andreisoare:
FYI this is how the table looks right now:

screenshot 2015-12-17 11 11 27

screenshot 2015-12-17 11 11 08

andreisoare:
And this is how tables are styled in Themis:

screenshot 2015-12-17 10 56 50

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,213 @@
...
+## Usage
+
+`th-table` receives a single attribute, the `delegate`, which is an object that
+manages the entire state of the table:
+* data
+* headers
+* pagination
+* sorting
+* loading and error states
+
```
JefCurtis: styling?

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,213 @@
...
+  * `updateData` is a callback that you can call with 3 optional parameters:
+    * `error` if there was an error while retrieving the data
+    * `data` which is an array of data objects
+    * `totalItems` which is used to compute the total number of pages
+
+* `headers` (optional)
+  * An array of TableHeader instances that define the name and sorting
+    capabilities of each table header.
+
+* `pageSize` (optional)
```
JefCurtis: What are you expecting here?

gulpfile.coffee
```diff
@@ -6,6 +6,11 @@ sass = require 'gulp-sass'
 rename = require 'gulp-rename'
 coffeelint = require 'gulp-coffeelint'
 
+autoprefixer = require 'gulp-autoprefixer'
+autoprefixerOptions =
+  browsers: ['last 2 versions']
```
seanhealy: Can we be a bit more explicit? With IE primarily I mean…

gulpfile.coffee
```diff
@@ -6,6 +6,11 @@ sass = require 'gulp-sass'
 rename = require 'gulp-rename'
 coffeelint = require 'gulp-coffeelint'
 
+autoprefixer = require 'gulp-autoprefixer'
```
seanhealy: Fuck yea!!!!!

polyfills/Array.find.polyfill.js
```diff
@@ -0,0 +1,22 @@
+if (!Array.prototype.find) {
```
seanhealy: I'm pretty sure you got this from the mozilla docs. Want to just add a line confirming this?

polyfills/String.includes.polyfill.js
```diff
@@ -0,0 +1,5 @@
+if (!String.prototype.includes) {
```
seanhealy: Same here.

polyfills/jQuery.typeAttribute.patch.js
```diff
@@ -4,7 +4,7 @@
 // https://github.com/jquery/jquery/commit/aad235b3251494afe71fd5bb6031e11965af9bdb
 
 if(typeof jQuery == 'function' || typeof $ == 'function') {
```
seanhealy: Is there any reason to check for `$` here? I'm pretty sure jQuery will always expose itself as `jQuery` but not always as `$` as `$` is sometimes used by other libs. Prototype for example which we use.

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
```diff
@@ -0,0 +1,57 @@
+fixtures = ->
```
seanhealy: Thoughts on putting the fixture after the angular controller? As it is secondary to the controller when a user is looking for an example.

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,227 @@
...
+</th-table-row>
+```
+
+This allows the developer to customize the message that's displayed when the
+table has no data to display.
+
+If not defined, it shows by default the following message: `No Results`.
+
+
+
```
JefCurtis: Highlight these three as services available to the implementer and that must be injected?

## Available Services for thTable

### SimpleTableDelegate

  **mandatory**
  _instructions here_
### TableHeader 

  **optional**
  _instructions here_
### TableSort 

  **optional: used with small datasets**
  _instructions here_

themis_components/thTable/examples/3 - Async Sort And Pagination/coffee.coffee
```diff
@@ -0,0 +1,55 @@
...
+          name: 'Last Name'
+          sortField: 'lastName'
+      ]
+
+      pageSize: 5
+
+      fetchData: ({currentPage, pageSize, sortHeader}, updateData) ->
+        sortedData = sort data, sortHeader
+        paginatedSortedData = getDataPage sortedData, currentPage, pageSize
+        setTimeout ->
```
seanhealy: You could import `$timeout` here and then you wouldn't need the manual digest trigger.

themis_components/thTable/examples/6 - Async Error/coffee.coffee
```diff
@@ -0,0 +1,30 @@
...
+      headers: [
+        TableHeader
+          name: 'Id'
+
+        TableHeader
+          name: 'Name'
+      ]
+
+      fetchData: (options, updateData) ->
+        setTimeout ->
```
seanhealy: `$timeout` here as well.

themis_components/thTable/index.scss
```diff
@@ -0,0 +1,172 @@
...
+    }
+  }
+
+  .th-table-overlay {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    z-index: 10;
```
seanhealy: Do we actually need the `z-index` here? It looks like it's logically ordered correctly so it might not be needed. I find it's best to avoid `z-index` if possible.

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,227 @@
+# DataTable — `thTable`
+
+## Description
+
+This component was built to replace all DataTables used in Clio.
```
seanhealy: It wouldn't be bad to talk about some of the motivations around `thTable` to help ease people into it. How it is lighter weight, more modular and such.

themis_components/thTable/readme.md
seanhealy: Agree, would be good to outline what is expected for these other fields. :)

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,227 @@
...
+  * When this is set, the table will display pagination controls at the bottom.
+
+* `currentPage` (optional, default value: 1)
+  * The current page to display when the table is initialized.
+
+
+The SimpleTableDelegate instance exposes the following methods and properties
+that you can use:
+
+* `headers`
```
seanhealy: Also perhaps what these return.

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,227 @@
...
+
+* `goToPrevPage()`
+  * Manually navigate to the previous page.
+
+* `goToPage(pageIndex)`
+  * Manually navigate to the specified page.
+
+
+
+## TableHeader
```
seanhealy: Thoughts on adding a table of contents to this doc? Just so people can get an outline and jump around?

gulpfile.coffee
andreisoare: We have quite too many actually. I don't know if I should clean them up in this ticket. How about creating a ticket to do this?

themis_components/thTable/thSimpleTableDelegate.service.coffee
```diff
@@ -0,0 +1,175 @@
+angular.module 'ThemisComponents'
+.factory 'SimpleTableDelegate', (TableDelegate, $interpolate) ->
+  interpolateStart = $interpolate.startSymbol()
+  interpolateEnd = $interpolate.endSymbol()
+
+  SimpleTableDelegate = (options = {}) ->
+    delegate = TableDelegate options
+
+    {
+      headers = []
```
seanhealy: I had to upgrade `coffeeify` in Themis to use this syntax. Have you seen any issue with it? We may want to update our package to explicitly have the correct version.

http://coffeescript.org/#changelog

polyfills/Array.find.polyfill.js
andreisoare: No, I worked 2 nights on this!

just kidding 😂

themis_components/thTable/thSimpleTableDelegate.service.coffee
```diff
@@ -0,0 +1,175 @@
...
+    generateCellsRowTemplate = (cellsRow, hasActionsRow) ->
+      ngRepeat = if hasActionsRow then "ng-repeat-start" else "ng-repeat"
+      borderBottomClass = "th-table-border-bottom" unless hasActionsRow
+      objectReference = getObjectReference cellsRow
+      cells = childrenArray(cellsRow)
+                .map (cell) -> generateCellTemplate cell
+                .join ''
+      return """
+        <tr class="th-table-cells-row #{borderBottomClass}"
+            #{ngRepeat}="#{objectReference} in thTable.delegate.getData()"
```
seanhealy: Should we mention in the docs that what ever `objectReference` is used for a row will carryover into the actions row as well.

themis_components/thTable/thSimpleTableDelegate.service.coffee
```diff
@@ -0,0 +1,175 @@
...
+      """
+
+    generateCellTemplate = (cell) ->
+      return """
+        <td>#{cell.innerHTML}</td>
+      """
+
+    generateActionsRowTemplate = (actionsRow, numColumns) ->
+      return "" unless actionsRow?
+      startColumn = parseInt(actionsRow.getAttribute('start-column')) || 1
```
seanhealy: We might want to check this with `?` as `0` is falsey.

```coffeescript
startColumn = parseInt(actionsRow.getAttribute('start-column')) ? 1
```

themis_components/thTable/thSimpleTableDelegate.service.coffee
```diff
@@ -0,0 +1,175 @@
...
+            ng-repeat-end
+            ng-mouseover="hover = true"
+            ng-mouseleave="hover = false"
+            ng-class="{'th-table-hover-row': hover}">
+          #{actions}
+        </tr>
+      """
+
+    generateNoDataRowTemplate = (noDataRow = {}, numColumns) ->
+      contents = noDataRow.innerHTML or "No Results"
```
seanhealy:
```coffeescript
contents = noDataRow.innerHTML ? "No Results"
```

themis_components/thTable/thSimpleTableDelegate.service.coffee
seanhealy: Thoughts on shortening this to what I have above?

polyfills/jQuery.typeAttribute.patch.js
andreisoare: I think you're right, there's no need for $ here. Do you want me to make the change in my PR?

themis_components/thTable/thTable.directive.coffee
```diff
@@ -0,0 +1,19 @@
...
+    controllerAs: 'thTable'
+    controller: -> return
+    compile: (element, attrs) ->
+      table = Table {element: element[0]}
+      table.clear()
+
+      post: (scope, element, attrs, controller) ->
+        table.setDelegate controller.delegate
+        template = table.generateTableTemplate()
+        childScope = scope.$parent.$new false, scope
```
seanhealy: We should probably leave a comment here as to why this was necessary.

themis_components/thTable/index.scss
andreisoare: Indeed, good call!

themis_components/thTable/thSimpleTableDelegate.service.coffee
andreisoare: I haven't had any problems, no. It seems the latest version of coffeeify is 2.0.1 - think I should test that?

themis_components/thTable/thSimpleTableDelegate.service.coffee
andreisoare: Yes, I wanted to document this but forgot :-(

themis_components/thTable/thSimpleTableDelegate.service.spec.coffee
```diff
@@ -0,0 +1,129 @@
+context = describe
+describe 'ThemisComponents: Service: thSimpleTableDelegate', ->
+  SimpleTableDelegate = TableHeader = null
+
+  createDOMElement = (template) ->
+    div = document.createElement 'div'
+    div.innerHTML = template
+    div.firstChild
```
seanhealy: This isn't problematic for your usage but this will only work if the template contains a single root node.

themis_components/thTable/thTable.service.coffee
```diff
@@ -0,0 +1,42 @@
...
+
+Table = (options = {}) ->
+  {
+    element
+  } = options
+
+  delegate = undefined
+
+  isProperlyDefinedRow = (row) ->
+    row.tagName == 'TH-TABLE-ROW' and row.hasAttribute 'type'
```
seanhealy: Mild preference to `is` over `==` here.

themis_components/thTable/thTable.service.coffee
```diff
@@ -0,0 +1,42 @@
...
+
+  isProperlyDefined = ->
+    return false if element.children.length is 0
+    for child in element.children
+      return false unless isProperlyDefinedRow child
+    true
```
seanhealy: You've used returns for the rest of this function and as it is more complicated than a simple expression I think an explicit return would be nice here.

themis_components/thTable/thTable.service.coffee
```diff
@@ -0,0 +1,42 @@
...
+  throw new Error "You must pass a DOM element to Table." unless element?
+
+  throw new Error "<th-table> not properly configured!" unless isProperlyDefined()
+
+  getRows = ->
+    rows = {}
+    for row in element.children
+      type = row.getAttribute 'type'
+      rows[type] = row
+    rows
```
seanhealy: Wanna have a short chat about functions and expressions. I've been thinking about distinctions here and would like a second thought. Could also have an effect on our style.

themis_components/thTable/thTable.service.coffee
```diff
@@ -0,0 +1,42 @@
...
+
+  return Object.freeze {
+    clear: -> element.removeChild element.firstChild while element.firstChild?
+
+    setDelegate: (newDelegate) -> delegate = newDelegate
+
+    generateTableTemplate: ->
+      throw new Error "You cannot generate a template " + \
+                      "before setting a delegate!" unless delegate?
+      delegate.generateTableTemplate rows
```
seanhealy: Explicit return here too?

themis_components/thTable/thTable.service.spec.coffee
```diff
@@ -0,0 +1,97 @@
+context = describe
+describe 'ThemisComponents: Service: thTable', ->
+  Table = SimpleTableDelegate = null
+
+  createDOMElement = (template) ->
```
seanhealy: Should we define this at a higher level (like compile directive) so it is shareable?

themis_components/thTable/thTableDelegate.service.coffee
```diff
@@ -0,0 +1,86 @@
...
+.factory 'TableDelegate', (TablePagination) ->
+  TableDelegate = (options = {}) ->
+    {
+      headers = []
+      currentPage
+      pageSize
+      fetchData
+    } = options
+
+    throw new Error "TableDelegate needs a fetchData property" unless fetchData?
```
seanhealy: We could be a bit more descriptive with this message. Fetch data actually needs to be a function correct?

themis_components/thTable/thTableDelegate.service.coffee
```diff
@@ -0,0 +1,86 @@
...
+    loading = false
+    error = false
+    currentSortHeader = headers.find (header) -> header.isSortActive()
+
+    triggerFetchData = ->
+      loading = true
+      error = false
+
+      updateData = (options = {}) ->
+        newError = options.error
```
seanhealy: Any reason why you're not destructuring this?

themis_components/thTable/thTableDelegate.service.spec.coffee
```diff
@@ -0,0 +1,141 @@
...
+    inject (_TableDelegate_, _TableHeader_) ->
+      TableDelegate = _TableDelegate_
+      TableHeader = _TableHeader_
+
+  it 'exists', ->
+    expect(TableDelegate?).toBe true
+
+  describe '#constructor', ->
+    it 'throws unless fetchData is defined', ->
+      expect(-> TableDelegate()).toThrow()
```
seanhealy: If you decide to tighten up the constraints on this above you'll likely want to do the same thing here.

themis_components/thTable/thTableHeader.service.coffee
```diff
@@ -0,0 +1,54 @@
+angular.module 'ThemisComponents'
+  .factory 'TableHeader', -> TableHeader
+
+TableHeader = (options = {}) ->
+  {
+    name = ''
+    sortField
+    sortActive
```
seanhealy: Should we default this to `no`?

themis_components/thTable/thTableHeader.service.coffee
```diff
@@ -0,0 +1,54 @@
...
+    align = "left"
+  } = options
+
+  sortDirection = "ascending" if sortActive and not sortDirection?
+
+  if sortActive? and not sortField?
+    throw new Error "you need to define sortField to enable sorting"
+
+  if sortDirection? and sortDirection not in ["ascending", "descending"]
+    throw new Error "sortDirection can be either ascending or descending"
```
seanhealy: What if I want to sort by Magic! :wink:

themis_components/thTable/thTableHeader.service.coffee
```diff
@@ -0,0 +1,54 @@
...
+    sortField
+    sortActive
+    sortDirection
+    align = "left"
+  } = options
+
+  sortDirection = "ascending" if sortActive and not sortDirection?
+
+  if sortActive? and not sortField?
+    throw new Error "you need to define sortField to enable sorting"
```
seanhealy: Make these error messages full sentences?

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
```diff
@@ -0,0 +1,56 @@
+angular.module 'thDemo', ['ThemisComponents']
```
doliveirakn: this file is named coffee.coffee. What is up with that?

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
```diff
@@ -0,0 +1,56 @@
...
+  .controller "DemoController", (SimpleTableDelegate, TableHeader, TableSort) ->
+    data = fixtures()
+    {sort} = TableSort
+
+    @tableDelegate = SimpleTableDelegate
+      headers: [
+        TableHeader
+          name: 'Id'
```
doliveirakn: Mentioned to Sean yesterday. This is fine for now, but eventually we are going to want to have internationalized strings as table headers and by fixing it in the JS code, we wouldn't be able to (at least not for multiple langauges)

This is a solvable problem, and one we don't need to tackle right now, but we should be mindful of approaches we can do. Maybe if got our translations into JS we could do something like
```
headers: [
  TableHeader
    name: -> I18n.translate("demo_table.headers.id")
    sortField: 'id'
]
```

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,319 @@
...
+
+    <th-table-cell>
+      {{item.lastName}}
+    </th-table-cell>
+  </th-table-row>
+</th-table>
+```
+
+```coffeescript
+controller: (SimpleTableDelegate, TableHeader, TableSort) ->
```
doliveirakn: I think we should have comments about `TableSort` since it is only really useful if we have the full universe of data. For the most part, this will never be the case due to pagination.

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,319 @@
...
+
+
+
+## thTableRow
+
+A row can be defined by adding a `th-table-row` component inside `th-table`.
+
+It accepts 2 attributes:
+* `type`, which is needed to differentiate between multiple row types.
+* `object-reference` (optional, default value: `item`), which defines the symbol
```
doliveirakn: nice. I was wondering if we could do this. It'll be much easier from a cognitive perspective if we can do `matter.foo` when interacting with matters instead of just `item`. 

Especially when there are multiple th tables on the same page. 
:+1:

themis_components/thTable/thSimpleTableDelegate.service.spec.coffee
andreisoare: I actually had a question about this - do you think I should make it global, like compileDirective? (and add a comment about supporting single root node)

themis_components/thTable/readme.md
```diff
@@ -0,0 +1,319 @@
...
+This service creates table header objects that you can pass to the delegate.
+
+```coffeescript
+header = TableHeader({name, sortField, sortActive, sortDirection})
+```
+
+* `name` (optional, default value: `""`) is what will be displayed as a column
+  title in the table.
+
+* `sortField` (optional) is the field in the data objects that sorting will be
```
doliveirakn: This doesn't actually do any sorting does it? That is actually taken care of by the `TableSort` service?

If not, then are the string (eg. `regularField`) arbitrary and the developer can do whatever they want with them?

eg:
```
if sortField == 'foo'
  # sort by bar
end
```

themis_components/thTable/thTable.service.coffee
andreisoare: We should formalize a rule for when to use explicit returns.

themis_components/thTable/thTable.service.spec.coffee
andreisoare: Ah, just saw this, after I asked the same thing above. On it.

themis_components/thTable/thTableDelegate.service.coffee
andreisoare: I can't write `{newError}` because that's not how the field is called and I can't write `{error}` because I don't want to override the error from the parent scope.

themis_components/thTable/thTableDelegate.service.spec.coffee
andreisoare: You mean to throw if fetchData is not a function?

themis_components/thTable/thTableHeader.service.coffee
andreisoare: I don't see why anyone would pass no here, but I agree it's better for readability to have it default to no. Do you actually want no here, or false? I don't get why coffeescript has duplicates for true and false.

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
andreisoare: It's just for consistency with all the other examples in the component library. I assume they were named like this because they represent the HTML and CoffeeScript tabs in each example block, see below:

screenshot 2015-12-18 10 36 27

screenshot 2015-12-18 10 38 00

themis_components/thTable/examples/1 - Actions Row Sort/coffee.coffee
andreisoare: I assume when we do implement translations there won't be any changes needed to the table.

themis_components/thTable/readme.md
andreisoare: If you use TableSort then they must be real fields in your objects. Otherwise, you can put whatever you want in sortField and that's what you will use to implement `fetchData({sortHeader}, updateData) ->`

Is this part not clear enough?

doliveirakn: Because I was reading the markdown and now through the UI, I didn't noticed the last bullet point (I thought there was only 3 points not 4). 

I think it is clear then.

themis_components/thTable/thTablePagination.service.coffee
```diff
@@ -0,0 +1,112 @@
...
+
+  TablePagination = (options = {}) ->
+    {
+      currentPage = 1
+      pageSize
+      triggerFetchData
+    } = options
+
+    totalItems = 0
+    ellipsis = '...'
```
seanhealy: We should use the ellipsis character here too as it will kern properly. `…` :)

themis_components/thTable/thSimpleTableDelegate.service.coffee
andreisoare: You were right, there were issues when installing lib-themisui in themis. Upgraded coffeeify and it's now fixed.

themis_components/thTable/thTableDelegate.service.coffee
seanhealy: `fetchData instanceOf Function`

themis_components/thTable/thTableDelegate.service.coffee
seanhealy: :+1:

PR #56, opened 2015-11-26
clio-27330 Fix disclosure overflow styling bug

andreisoare:


themis_components/thDisclosure/thDisclosureContent.directive.coffee
```diff
@@ -30,11 +30,14 @@ angular.module 'ThemisComponents'
...
-            $($element).css 'height', 'auto'
+            $($element).css
+              height: 'auto'
+              overflow: 'visible'
         else
           $($element).animate {
             height: "0"
-          }, 300
+          }, 300, ->
+            $($element).css 'overflow', 'hidden'
```
seanhealy: I would slightly prefer to use the `overflow: 'hidden'` syntax here. It reads nicer and is a bit more consistent.

PR #58, opened 2015-11-27
Clio-27373 - Fix scroll top bug in ThemisUI

andreisoare:


themis_components/thComponentExample/thComponentExample.directive.coffee
```diff
@@ -75,8 +75,11 @@ angular.module('ThemisComponents')
 
           exampleFrame.contentWindow.document.addEventListener "DOMSubtreeModified", ->
             $timeout ->
+              scrollingElement = exampleFrame.ownerDocument.scrollingElement.getElementsByClassName('component-details-view')[0]
```
seanhealy: Nitpicky but we could use `querySelector('.component-details-view')` here then we would also not need the `[0]`.

PR #59, opened 2015-11-30
clio-27344 Add the ability to pass local variables into modals.

JefCurtis:
- Add example of this.
- Updated tests and readme file.
- Update to thModal examples to use themis components instead of native HTML element. ex. `<thInput>` instead of `<input>`

### DIFF
--- a/public/exampleTemplates/thModalExampleTemplate.html
+++ b/public/exampleTemplates/thModalExampleTemplate.html
@@ -8,6 +8,6 @@
     proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
   </p>
   <div class='actions'>
-    <button ng-click="modal.dismiss()">close</button>
+    <th-button ng-click="modal.dismiss()">close</th-button>
   </div>
 </div>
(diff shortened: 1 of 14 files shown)


public/exampleTemplates/thModalExampleTemplate.html
```diff
@@ -8,6 +8,6 @@
     proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
   </p>
   <div class='actions'>
-    <button ng-click="modal.dismiss()">close</button>
+    <th-button ng-click="modal.dismiss()">close</th-button>
```
seanhealy: Nice cleanup here.

public/exampleTemplates/thModalExampleTemplate3.html
```diff
@@ -1,9 +1,6 @@
-<div style="text-align: center">
-  <div class="actions">
-    <label for="name">Username: </label>
-    <input type="text" name="name" ng-model="userName">
-    <br>
-    <button ng-click="modal.confirm(userName)">Submit</button>
-    <button ng-click="modal.dismiss()">Cancel</button>
-
+<div class="actions">
+  <th-input with-label="Name:" type="text" name="name" ng-model="name">
```
seanhealy: :+1:

PR #60, opened 2015-12-01
Clio 27406 - Add coffeelint

andreisoare:


coffeelint.json
```diff
@@ -0,0 +1,55 @@
...
+  },
+
+  "line_endings": {
+    "value": "unix",
+    "level": "error"
+  },
+
+  "max_line_length": {
+    "name": "max_line_length",
+    "value": 100,
```
seanhealy: It might be nice for us to slowly reduce this. This is a good starting point but always something we can come back to reduce and cleanup.

coffeelint.json
```diff
@@ -0,0 +1,55 @@
...
+    "level": "error",
+    "limitComments": true
+  },
+
```
seanhealy: Do you think we might want to add `missing_fat_arrows`? Thoughts?

coffeelint.json
```diff
@@ -0,0 +1,55 @@
...
+  "no_stand_alone_at": {
+    "level": "error"
+  },
+
+  "no_this": {
+    "level": "error"
+  },
+
+  "space_operators": {
+    "level": "error"
```
seanhealy: Does `!` count as an operator in this case?

andreisoare: No.

coffeelint.json
andreisoare: +1

coffeelint.json
andreisoare: missing_fat_arrows produces annoying false positives. While useful, I feel it will complain too much.

no_debugger is enabled by default (level warn).

Added no_empty_functions.

karma.conf.coffee
```diff
@@ -17,11 +17,10 @@ module.exports = (config) ->
         'http://code.jquery.com/jquery-2.1.4.js'
         'index.coffee'
         'node_modules/angular-mocks/angular-mocks.js'
-        'themis_components/**/*.mock.coffee'
         'themis_components/**/*.spec.coffee'
         'spec_helper/compileDirective.coffee'
         {
-            pattern: 'themis_components/**/*.{directive|service}.coffee'
+            pattern: 'themis_components/**/*.+(directive|service).coffee'
```
seanhealy: Not a problem but wondering why you made this change…?

karma.conf.coffee
```diff
@@ -38,6 +37,7 @@ module.exports = (config) ->
     # available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
     preprocessors: {
         'index.coffee' : ['browserify']
+        'themis_components/**/*.coffee': ['coffeelint']
```
seanhealy: Might also be valuable to lint our tests…?

public/javascript/controllers/AppController.coffee
```diff
@@ -1,4 +1,4 @@
 angular = require 'angular'
 
 angular.module('ThemisComponentsApp')
-  .controller 'AppController', ->
+  .controller 'AppController', -> undefined
```
seanhealy: thoughts on an empty `return` here instead of `undefined`. Since we need the hanging return on controllers anyway it might fit a bit better…

themis_components/thDisclosure/thDisclosureManager.service.coffee
```diff
@@ -4,7 +4,9 @@ angular.module 'ThemisComponents'
 
     onToggle = (name, handler) -> handlerMap[name] = handler
 
-    toggle = (name) -> (handlerMap[name] ? ->)()
+    toggle = (name) -> (handlerMap[name] ? -> undefined)()
```
seanhealy: Same thought re `return` here.

themis_components/thModal/thModal.directive.spec.coffee
```diff
@@ -1,7 +1,7 @@
 context = describe
 describe "ThemisComponents: Directive: thModal", ->
   directive = element = scope = ModalManager = null
-  beforeEach ->
+  beforeEach -> undefined
```
seanhealy: `return`?

public/javascript/controllers/AppController.coffee
andreisoare: Oh yes, good idea!

karma.conf.coffee
andreisoare: Right now on master if you run `karma start` you will get this message:

```
08 12 2015 09:54:32.015:WARN [watcher]: Pattern "/Users/Andrei/Work/lib-themisui/themis_components/**/*.{directive|service}.coffee" does not match any file.
```

Because this pattern doesn't work, it also prevents preprocessors on this file pattern, so coffeelint doesn't work.

PR #61, opened 2015-12-01
Clio 26873 thcheckbox

dtomalty:
Sorry for the messy diff on specs. I reworked them to use the existing compileDirective function.


themis_components/thCheckbox/index.scss
```diff
@@ -43,4 +43,8 @@
       color: $dark-grey-4;
     }
   }
+
+  input {
```
seanhealy: Not a huge deal but it would be a bit more specific to use `input[type="checkbox"] {` here.

themis_components/thCheckbox/readme.md
```diff
@@ -8,5 +8,5 @@ You use `thCheck` create and manage checkboxes used, for example, in forms. This
 
 ### Markup
 ```
-<th-checkbox ng-model="state"></th-checkbox>
+<th-checkbox ng-model="state" ng-change="someAction()"></th-checkbox>
```
seanhealy: Would be good to expand this read me a bit explaining `ng-change` as it's not a required attribute which this could make a bit confusing.

themis_components/thSwitch/readme.md
```diff
@@ -8,5 +8,5 @@ You use `thSwitch` create and manage the On/Off buttons used, for example, in th
 
 ### Markup
 ```
-<th-switch ng-model="state"></th-switch>
+<th-switch ng-model="state" ng-change="someAction()"></th-switch>
```
seanhealy: Same here.

themis_components/thCheckbox/thCheckbox.directive.coffee
```diff
@@ -4,13 +4,20 @@ angular.module('ThemisComponents')
...
-  @checked = @checked ? false
-  @toggle = -> @checked = not @checked
+      @toggle = ->
+        $scope.$apply =>
+          @checked = not @checked
+        @change() if @change?
 
-  return
+      $element.on 'click', =>
+        @toggle()
```
seanhealy: It would be slightly more proper to use the `$scope.$apply` in the event here as the event is why it's needed.

themis_components/thSwitch/index.scss
```diff
@@ -19,6 +19,10 @@
   overflow: hidden;
   cursor: pointer;
 
+  input {
```
seanhealy: Same here.

themis_components/thSwitch/thSwitch.directive.coffee
```diff
@@ -4,11 +4,20 @@ angular.module('ThemisComponents')
...
       state: '=ngModel'
     bindToController: true
     controllerAs: 'switch'
-    controller: ->
+    controller: ($scope, $element) ->
       @state = @state ? off
-      @toggle = -> @state = not @state
+
+      @toggle = ->
+        $scope.$apply =>
```
seanhealy: Same here.

themis_components/thCheckbox/thCheckbox.directive.spec.coffee
```diff
@@ -1,47 +1,55 @@
...
+
+    directive = compileDirective(template, scopeAdditions)
+    element = directive.element
+    scope = directive.scope
+
+  describe 'when checkbox is unchecked', ->
+    beforeEach ->
+      compileCheckboxDirective(validTemplate)
+
+    it 'should have an i element', ->
```
seanhealy: This is really nitpicky and not a change you made but it would be more apt to verify that it has an indicator than an `i` tag.

themis_components/thCheckbox/thCheckbox.directive.coffee
dtomalty: I use `$scope.$apply` on `@checked` only because otherwise the updated value is not reflected in the `ng-change` expression that is evaluated on the next line. If a user were to check the value of `@checked` in the callback it would be the old value.

themis_components/thCheckbox/readme.md
```diff
@@ -10,3 +10,7 @@ You use `thCheck` create and manage checkboxes used, for example, in forms. This
 ```
 <th-checkbox ng-model="state"></th-checkbox>
 ```
+Specify an optional expression to evaluate using `ng-change` and/or `ng-click`.
```
PANCAKES: Can you use both ```ng-click``` and ```ng-change``` at the same time? Which one is applied first, if so?

themis_components/thSwitch/readme.md
```diff
@@ -10,3 +10,7 @@ You use `thSwitch` create and manage the On/Off buttons used, for example, in th
 ```
 <th-switch ng-model="state"></th-switch>
 ```
+Specify an optional expression to evaluate using `ng-change` and/or `ng-click`.
+```
+<th-switch ng-model="state" ng-change="someAction()"></th-switch>
```
PANCAKES: Same comment about ```ng-click``` and ```ng-change``` being used at once

themis_components/thCheckbox/readme.md
dtomalty: Yes you can use both. Change is applied first, which is consistent with how the normal checkbox works.

PR #65, opened 2015-12-09
clio-23323 th-lazy improvements to error handling

JefCurtis:
- Add an error message of load failure. This message can be customized. 
- Add test to incorporate the changes.


themis_components/thLazy/readme.md
```diff
@@ -4,10 +4,13 @@
 
 If you want to lazy load some content `th-lazy` is the directive for you. It is a pretty simple wrapper around `ng-include` with the addition of a visual indicator that something is loading.
 
-`th-lazy` accepts a `src` URL of where it will fetch its inner content from.
+`th-lazy` accepts a:
+- `src` URL of where it will fetch its inner content from.
+- `error-message` a message to be displayed if there is an error loading the template. _if_ nothing is passed in then defaults to "There was a problem loading your page"
```
seanhealy: That error message is very technical. We should consider something that is a bit more user friendly. It's generally a good idea for errors to provide users with a path forwards rather than a description of what actually happened. 

http://uxmovement.com/forms/how-to-make-your-form-error-messages-more-reassuring/

Off hand something like this would be a bit better.
"Missing View. Try reloading or, contact support."

themis_components/thLazy/examples/1 - Simple/html.html
```diff
@@ -1,7 +1,5 @@
 <html ng-app="thDemo">
   <body>
-
-    <th-lazy src="exampleTemplates/thLazyExampleTemplate.html"></th-lazy>
-
+    <th-lazy src="exampleTemplates/thLazyExampleTemplate.html" error-message="Sorry! Can't find what you are looking for"></th-lazy>
```
seanhealy: Since the `error-message` isn't required it probably isn't necessary to show it in the simple example.

themis_components/thLazy/readme.md
JefCurtis: Thanks for the link. I'll read it through and use your suggestion. This should probably go through the design team anyways.

themis_components/thLazy/examples/1 - Simple/html.html
JefCurtis: Removed.

themis_components/thLazy/index.scss
```diff
@@ -1,4 +1,19 @@
-.th-lazy-loader {
-  padding: 1rem;
+.th-error-container {
```
seanhealy: Sorry, last minute tis here. We should nest both of these under `th-lazy` so that they don't leak.

PR #66, opened 2015-12-14
Clio 23455 thRadioGroup/thRadioButton components

dtomalty:


themis_components/thRadioGroup/thRadioGroup.directive.spec.coffee
```diff
@@ -0,0 +1,138 @@
...
+      value: value
+      callback: callback
+
+    directive = compileDirective(template, scopeAdditions)
+    element = directive.element
+    scope = directive.scope
+
+  describe 'when value is undefined', ->
+    beforeEach ->
+      compileRadioGroupDirective(validTemplate, undefined)
```
seanhealy: In JS if you don't need a later parameter you can just ignore it so the following is also valid and probably a bit more common to be seen.

```coffeescript
compileRadioGroupDirective validTemplate
```

themis_components/thRadioGroup/thRadioGroup.directive.spec.coffee
```diff
@@ -0,0 +1,138 @@
...
+
+    it 'should have an indicator for each button', ->
+      expect(element.find('i').length).toBe 2
```
seanhealy: It might be better to move down to an individual radio button and verify that it has one `i`. That way we are less sensitive to the template.

themis_components/thRadioGroup/thRadioGroup.directive.spec.coffee
```diff
@@ -0,0 +1,138 @@
...
+
+    it 'should be unchecked', ->
+      expect(element.find('.th-radio-button').first().hasClass('checked')).toBe false
```
seanhealy: This is a hard one but Angular 2 no longer requires jQuery so we should try to lean on it less. Not a big deal but it would be a bit better to convert these selectors to native JS.

Along these lines:
```coffeescript
element.querySelector(.th-radio-button:first-child).matches('.checked')
```

themis_components/thRadioGroup/thRadioGroup.directive.spec.coffee
```diff
@@ -0,0 +1,138 @@
...
+      expect(element.find('.th-radio-button').last().hasClass('checked')).toBe false
+
+    describe 'when first element is clicked', ->
+      beforeEach ->
+        element.find('.th-radio-button').first().triggerHandler 'click'
+
+      it 'should have second element checked only', ->
```
seanhealy: should have first element checked only

themis_components/thRadioGroup/thRadioGroup.directive.spec.coffee
seanhealy: It might be useful to build a function in your tests here that gets you the selected state of a radio button dom element you pass to it. You're replicating this code in a lot of places.

PR #67, opened 2015-12-16
clio-27937 th-error

JefCurtis:
- Add `th-error` component with tests. 
- Replace `th-lazy`'s error handling with `th-error`.


themis_components/thError/thError.directive.coffee
```diff
@@ -0,0 +1,21 @@
+angular.module("ThemisComponents")
+  .directive "thError", ->
+    restrict: "E"
+    transclude: true
+    template: """
```
PANCAKES: Were we moving the templates to their own file or are we not enforcing that yet?

JefCurtis: No problem, I can move it.

themis_components/thError/index.scss
```diff
@@ -0,0 +1,21 @@
+.th-error-container {
```
andreisoare: You should remove the CSS from `th-lazy`, otherwise we will duplicate this entire CSS.

JefCurtis: Done. thanks!

themis_components/thError/readme.md
```diff
@@ -0,0 +1,16 @@
...
+
+### Markup
+```
+<th-error></th-error>
+```
+
+Pass in transcluded elements (will be effected by the same styling as default message)
+
+```
+<th-error><p>transcluded message</p></th-error>
```
seanhealy: Do I need a `<p>` around my error message?

Also, the word transcluded here is a bit unnecessarily technical and speaks to us more than to the function of the component.

JefCurtis: updated to make it (hopefully) clearer.

PR #71, opened 2015-12-23
clio-28507 Uglify lib-themisui.min.js. Remove unnecessary packages.

andreisoare:


package.json
```diff
@@ -47,11 +47,9 @@
   },
   "devDependencies": {
     "angular-mocks": "1.4.7",
-    "brfs": "^1.4.0",
     "jasmine-core": "^2.3.4",
     "karma": "^0.13.15",
     "karma-browserify": "^4.1.2",
-    "karma-cli": "^0.1.1",
```
andreisoare: This really is not needed, I removed it. Karma still works because of the karma package. NPM knows where to find the binary because of [this entry in karma's package.json](https://github.com/karma-runner/karma/blob/d60ac8b8a67a567dd1b87628aeac0b1443685e88/package.json#L324-L326).

PR #72, opened 2015-12-30
clio-28939 Fix thTable loading margin-bottom

andreisoare:


README.md
```diff
@@ -12,7 +12,7 @@ ThemisUI is a shareable suite of components and brand assets to be shared across
 We're running in stealth mode for now meaning we don't yet want to list on NPM. So to install run:
 
 ```bash
-npm install git+ssh://git@github.com/clio/lib-themisui.git
+npm install clio/lib-themisui
```
seanhealy: Good call.

themis_components/thTable/index.scss
```diff
@@ -6,6 +6,10 @@ th-table {
 
   .th-table-loading {
     opacity: 0.5;
+
+    &.th-table-blank table.th-table {
+      margin-bottom: 62px;
```
seanhealy: Where is this magic number coming from?

themis_components/thTable/index.scss
```diff
@@ -13,6 +17,7 @@ th-table {
     border-collapse: collapse;
     width: 100%;
     font-size: inherit;
+    margin-bottom: 16px;
```
seanhealy: And here.

andreisoare: This is the default margin-bottom for tables in Themis.

themis_components/thTable/index.scss
andreisoare: I want to keep the same 16px below the loading spinner.

The spinner has margin-top 50 from the beginning of the table and height 30, so it ends at 80.

The table already has the header of 35 height, so we need to add another 80 + 16 - 35 = 61.

So it should actually be 61, not 62. Let me know if this math makes sense and if I should change it to 61.

seanhealy: That makes sense. Could we build some of these values off of variables and or add a comment explaining the value.

themis_components/thTable/index.scss
seanhealy: Do we have a variable that we could use for this?

themis_components/thTable/index.scss
```diff
@@ -1,4 +1,8 @@
 th-table {
+  $th-table-cell-height: 35px;
+  $th-table-margin-bottom: 16px;
+  $th-table-loading-spinner-height: 30px;
```
andreisoare: Unfortunately this cannot be controlled inside th-table.

PR #78, opened 2016-01-11
CLIO-29377: attach type 'button' to all buttons by default. This is p…

dtomalty:
…art of the HTML5 spec and prevents them from submitting in forms. Update spec.


themis_components/thButton/thButton.directive.spec.coffee
```diff
@@ -25,8 +25,8 @@ describe 'ThemisComponents: Directive: thButton', ->
   it 'has the button text set correctly', ->
     expect(element.text().trim()).toBe 'some text'
 
-  it 'has no button type', ->
-    expect(element.attr('type')).toBe undefined
+  it 'has type "button"', ->
```
andreisoare: Can you also make a test that fails without the changes in this PR? Basically create a form with the issue that we had in Themis.

themis_components/thButton/thButton.directive.coffee
```diff
@@ -15,6 +15,7 @@ angular.module('ThemisComponents')
     bindToController: true
     controllerAs: 'button'
     controller: ($element, $attrs) ->
-      $element.attr('type', 'submit') if $attrs.submit?
+      type = if $attrs.submit? || $attrs.type == 'submit' then 'submit' else 'button'
```
andreisoare: Coffeescript likes `or` and `is` over default JS operators. Also storing `$attrs.submit? || $attrs.type == 'submit'` in a variable with a clear name will make the code more readable.

themis_components/thButton/thButton.directive.spec.coffee
andreisoare: Actually just a comment that explains the behaviour will be fine.

PR #81, opened 2016-01-13
CLIO-29267: Alert component

craigcarlyle:
* Added thAlert


themis_components/thAlert/readme.md
```diff
@@ -0,0 +1,48 @@
...
+or for an announcement.
+
+## Usage
+
+`<th-alert-anchor></th-alert-anchor>` is required in the body of your application.
+It's positioned absolutely to appear at the top of the page, so ensure it's not
+placed in a relatively positioned container.
+
+There are 3 types of alerts that are called with their own functions
+using the `AlertManager`:
```
andreisoare: I would specify that `AlertManager` is a service.

`using the AlertManager service:`

themis_components/thAlert/thAlertAnchor.directive.coffee
```diff
@@ -0,0 +1,12 @@
+angular.module('ThemisComponents').directive "thAlertAnchor", ->
+  restrict: "E"
+  bindToController: true
+  controllerAs: 'alert'
```
andreisoare: We name controllers just like components, so: `controllerAs: 'alertAnchor'`.

themis_components/thAlert/index.scss
```diff
@@ -0,0 +1,64 @@
...
+  left: 0;
+  width: 100%;
+  color: #fff;
+  z-index: 2147483647;
+  height: 54px;
+  line-height: 54px;
+  padding: 0 16px;
+  display: block;
+  @include animated(200ms);
+  @include box-sizing;
```
seanhealy: What does box-sizing do here?
We have auto prefixer in ThemisUI so we don't need browser prefix fixes here.

themis_components/thAlert/index.scss
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    &.fa-check {
+      background: $green-5;
+    }
+
+    &.fa-times-circle {
+      cursor: pointer;
+      line-height: 54px;
+      float: right;
```
seanhealy: Is this to position a close icon?

We should get out of the habit of using float for positioning. It's not meant for this and often causes unforeseen issues.
Given what's going on in here it would be better to absolutely position this or use flex box.

themis_components/thAlert/index.scss
```diff
@@ -0,0 +1,64 @@
+.th-alert {
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100%;
+  color: #fff;
+  z-index: 2147483647;
+  height: 54px;
+  line-height: 54px;
+  padding: 0 16px;
```
seanhealy: We shouldn't use short hand notion here.

http://codeguide.co/#css-shorthand

PR #89, opened 2016-02-02
Clio-30008 thFilter Design Proposal

andreisoare:


themis_components/thFilter/readme.md
```diff
@@ -0,0 +1,172 @@
...
+    without having to click the "Add" button and enable it from a list.
+
+### Autocomplete Filter
+
+Uses our `thAutocomplete` component to query the server for options as you type.
+
+It requires passing 3 more options through the `autocompleteOptions` property.
+These 3 options must be supported by `thAutocomplete` as well.
+
+* `ModelClass` (string)
```
andreisoare: The only worry I have here is that if we're serving these from the backend, then it will have to know what model classes we have on the frontend :cry:

andreisoare: After thinking more about this I think we can get rid of the autocomplete filter for now, as it's only used for matters and contacts. We could create a `ccMattersAutocomplete` directive and a `ccContactsAutocomplete` directive and use them as filter types.

themis_components/thFilter/readme.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+```coffeescript
+controller: ->
+  # In reality @filters would actually be retrieved with an AJAX call, in order
+  # to support custom fields.
+  #
+  # I'm harcoding it here to get a sense of all the possible types of fields and
+  # how they should be configured.
+  @filters = [
```
seanhealy: I'm thinking we might want to differentiate between available filters and applied filters. So, one array for filters that can be chosen then a second for filters that the user has chosen / activated.

Also, do we want to make a consideration that a user may want to use the same filter twice. Such as this?

themis_components/thFilter/readme.md
seanhealy: This is an 'interesting' consideration. I'm having that feeling that tells me there is a nice solution around it but need to think a bit more to see if it's just by gut tricking me. :p

themis_components/thFilter/readme.md
```diff
@@ -0,0 +1,172 @@
...
+## Filter object
+
+All field objects support the following properties:
+
+* `fieldName` (string)
+  * Useful to pass to th-table to know what field to request from the API.
+* `label` (string)
+  * Useful to show to the user what this filter field actually is.
+* `type` (string)
+  * One of: "autocomplete", "select", "search", "date", "number".
```
seanhealy: `search`, `date` and `number` are all pretty straight forward here.

You've put some thought into `autocomplete` below but, I think we need to consider select a bit more explicitly as well. Do we want just an array of keys, a key, value array…

themis_components/thFilter/readme.md
andreisoare: Yes, this ties in with the `onSave()` function that's planned for the future. Good catch.

andreisoare: I don't think having the same filter twice will be needed as long as we implement proper operators. In your case it would be "Last opened date" is between <date1> and <date2>.

For strings, for example, you can support things like: "Name" contains any of the words <words>.

themis_components/thFilter/readme.md
andreisoare: Good call with the (k,v) array.

andreisoare: :white_check_mark: Resolved

themis_components/thFilter/readme.md
```diff
@@ -0,0 +1,238 @@
...
+```coffeescript
+controller: ->
+  # In reality @filters would actually be retrieved with an AJAX call, in order
+  # to support custom fields.
+  #
+  # I'm harcoding it here to get a sense of all the possible types of fields and
+  # how they should be configured.
+  @filters = [
+    {
+      fieldName: "attorney"
```
seanhealy: Thoughts on calling this `fieldIdentifier` or just `identifier` as this is really a unique key that is used to identify this filter. We are creating a DDB sort of structure here.

themis_components/thFilter/readme.md
```diff
@@ -0,0 +1,238 @@
...
+  ]
+
+  @onFilterChange = (filters) =>
+    # Called every time a filter is changed. This is an opportunity to reload
+    # the table, assuming that the table's fetchData uses the @tableFilters
+    # field set on the controller when making requests.
+    @tableFilters = filters
+    @tableDelegate.reload {currentPage: 1}
+
+  @saveFilters: (filters) =>
```
seanhealy: Should we have a `loadFilters` as well?

Or, perhaps just a `serialize` and `unserialize`…

themis_components/thFilter/readme.md
```diff
@@ -0,0 +1,238 @@
...
+
+All field objects support the following properties:
+
+* `fieldName` (string)
+  * Useful to pass to th-table to know what field to request from the API.
+* `label` (string)
+  * Useful to show to the user what this filter field actually is.
+* `type` (string)
+  * One of: "autocomplete", "select", "string", "date", "number".
+* `showOnInit` (boolean, default: false)
```
seanhealy: I'm guessing we will want this but we should confirm this with design before implementing this option.

themis_components/thFilter/readme.md
```diff
@@ -0,0 +1,238 @@
...
+
+`thFilter` will make the request and parse the response internally. It uses
+`labelField` to show a specific field for each object from the response array.
+
+### Select filter
+
+Requires an additional `data` field (array) which contains all the options that
+the user can select.
+
+The `data` array is actually an array of `{label, value}` objects: the label is
```
seanhealy: Do you mean here an array of objects with keys `label`, `value` or more of an tuple `[[text, value], …]`?

themis_components/thFilter/readme.md
```diff
@@ -0,0 +1,238 @@
...
+This is a set of components and services that you can use to implement table
+filters in Clio.
+
+It supports custom fields that have to be populated by making a network request
+to the backend.
+
+## Example
+
+```html
+<th-filter filters="controller.filters"
```
seanhealy: Nitpicky but we should do one attribute per line.

```html
<th-filter
  filters="controller.filters"
  activeFilters="controller.activeFilters"
  onFilterChange="controller.onFilterChange"
  saveFilters="controller.saveFilters"
  >
</th-filter>
```

themis_components/thFilter/readme.md
```diff
@@ -0,0 +1,238 @@
...
+           activeFilters="controller.activeFilters"
+           onFilterChange="controller.onFilterChange"
```
seanhealy: Thoughts on doing a delegate again here for these callbacks?

themis_components/thFilter/readme.md
andreisoare: I would leave this as it is with `ModelClass`. The backend doesn't actually need to be aware of the frontend models.

Right now it sends filter types to the client, like: matterAutocomplete.  In the frontend we could process that to the proper `ModelClass`, `labelField` and `valueField`.

themis_components/thFilter/readme.md
andreisoare: :white_check_mark: Resolved

themis_components/thFilter/readme.md
andreisoare: :white_check_mark: Resolved

themis_components/thFilter/readme.md
andreisoare: :white_check_mark: Resolved

themis_components/thFilter/readme.md
andreisoare: :white_check_mark: Resolved

themis_components/thFilter/readme.md
andreisoare: :white_check_mark: Resolved

themis_components/thFilter/readme.md
andreisoare: Geoff wasn't sure about it, he said he'd prefer to have this option, but didn't yet know how it would look like. I'll leave it here for now, can be easily removed.

themis_components/thFilter/readme.md
andreisoare: :white_check_mark: Resolved

themis_components/thFilter/readme.md
seanhealy: :white_check_mark: Resolved

themis_components/thFilter/readme.md
seanhealy: :white_check_mark: Resolved

PR #97, opened 2016-02-20
quick fix for CSS selector problem

JefCurtis:
There seems to be a problem using the `:not()` selector. Replacing this with a hardcoded call to the specific classes seems to fix this.

### DIFF
--- a/themis_components/thAlert/index.scss
+++ b/themis_components/thAlert/index.scss
@@ -50,7 +50,7 @@
     vertical-align: middle;
     font-size: 18px;
 
-    &.fa-check, &.fa-warning {
+    &.alert-icon {
       color: #fff;
       text-align: center;
       width: 40px;
@@ -64,7 +64,7 @@
       background: $green-5;
     }
 
-    &.fa-times-circle {
+    &.dismiss-icon {
       cursor: pointer;
       line-height: 54px;
       width: 16px;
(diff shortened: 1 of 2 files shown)


themis_components/thAlert/index.scss
```diff
@@ -50,7 +50,7 @@
     vertical-align: middle;
     font-size: 18px;
 
-    &:not(.fa-times-circle) {
+    &.fa-check, &.fa-warning {
```
seanhealy: I assume we're using [`fa-times-circle`](http://fontawesome.io/icon/times-circle/) for our icon if there is an error?

Could you look at how hard it would be to instead switch that to [`times`](http://fontawesome.io/icon/times/)? It would allow us to treat all the icons in the same way instead of making special cases like we have now.

JefCurtis: Sure I'll check it out today. Good idea.

JefCurtis: After give this a look the `fa-times-circle` class is actually being used at the dismiss icon on the right side. It needs it's own styling.

seanhealy: Ah… Didn't realize that.

In that case I thing I would instead make a class to identify the close button so we can make this exemption in a more clear manner.

PR #100, opened 2016-02-29
CLIO-24155: Popover improvements.

dtomalty:


themis_components/thPopover/examples/4 - Trigger from script/coffee.coffee
```diff
@@ -0,0 +1,9 @@
+angular.module('thDemo', ['ThemisComponents'])
+  .controller 'DemoCtrl', (PopoverManager) ->
+    @clickHandler = ->
+      PopoverManager.showPopover(
+        'target'
+        PopoverManager.getContent('content')
+      )
```
seanhealy: This parentheses are unneeded.

themis_components/thPopover/examples/3 - Inline content/html.html
```diff
@@ -0,0 +1,14 @@
+<html ng-app="thDemo">
+  <body style="height: 200px;">
+
+    <th-button th-popover="test">Trigger</th-button>
+
+    <th-popover-content name="test">
+      <div>
+        <h1>thPopover Example 3</h1>
+        <div><a href="" ng-click="dismiss()">Close</a></div>
```
seanhealy: Do we need the `dismiss` function here?

themis_components/thPopover/readme.md
```diff
@@ -1,12 +1,18 @@
...
 
 Add small overlays of content, like those on the iPad, to any element for housing secondary information.
 
-`th-popover` accepts a URL of where it will fetch its inner content from.
+`th-popover` accepts the name of a `th-popover-content` element where its content will come from.
 
-`th-popover` also accepts an optional `overflow` param that will apply a styling class. This is intended to be used when dealing with legacy components that have elements that are positioned absolutely (such as datepickers). The overflow property is set to `auto` by default. Valid options:
+`th-popover-url` accepts a URL from which it will fetch its content.
+
+`th-popover-content` specifies a block of content used for display in a popover.
```
seanhealy: For this and for `th-popover-target` we should specify that these should be unique names.

themis_components/thPopover/thPopoverManager.service.coffee
```diff
@@ -0,0 +1,177 @@
...
+    attachPopover = ($scope, element, attributes, contentPromise) ->
+      {renderPopover} = addPopoverToTarget(
+        {$scope, element, attributes}
+        contentPromise
+      )
+
+      element.on 'click', ->
+        renderPopover()
+
+    addPopoverToTarget = (target, contentPromise) ->
```
seanhealy: Seeing this in practice I think it might be nicer for this logic to be included with a `require`. it's what is doing the DOM manipulation so it feels like a natural split so it would keep this file a bit more single purpose and smaller.

themis_components/thPopover/thPopover.directive.coffee
```diff
@@ -1,121 +1,13 @@
...
-            $scope.dismiss()
-
-
-      $scope.dismiss = ->
-        $scope.$broadcast 'thPopover.dismiss'
+      unless attributes.thPopover?.length > 0
+        throw new Error "thPopover: must specify content name."
+
+      PopoverManager.attachPopover(
+        $scope.$new()
```
seanhealy: Why do we need a new scope here?

dtomalty: This happens in the `thPopover`, `thPopoverTarget` and `thPopoverUrl` directives. It's because `addPopoverToTarget` (the main DOM-manipulation function in `PopoverManager`) adds variables directly to the scope that would get overridden otherwise when more than one popover is initialized at the same time.

themis_components/thPopover/thPopoverManager.service.coffee
dtomalty: :white_check_mark: Resolved

dtomalty: Agreed.

themis_components/thPopover/thPopover.directive.coffee
dtomalty: I left this as is, for now. The templates are shared among the three popover directives and it would require having all directives use the same `controllerAs` name to bind to when compiling.

themis_components/thPopover/thAddPopover.helper.coffee
```diff
@@ -0,0 +1,127 @@
...
+
+      if bottom?
+        view.css bottom: "#{ bottom }px"
+
+      # Position the arrow
+      arrow.css
+        top: "#{ top }px"
+        left: "#{ anchorRect.left + anchorRect.width / 2 - arrowOffset }px"
+
+      overlay.removeClass 'th-popover-hidden'
```
seanhealy: Other than these classes did you have to do any other changes to the way the DOM was handled?

themis_components/thPopover/thPopover.directive.coffee
seanhealy: I was actually thinking about this a *bit* more. Would we be able to have `PopoverManager` generate this new scope off of `rootScope` so we don't need this parameter or is the scope inheritance important here?

themis_components/thPopover/examples/4 - Trigger from script/coffee.coffee
```diff
@@ -0,0 +1,9 @@
+angular.module('thDemo', ['ThemisComponents'])
+  .controller 'DemoCtrl', (PopoverManager) ->
+    @clickHandler = ->
+      PopoverManager.showPopover(
+        'target'
```
seanhealy: Thoughts on using named params here?

```coffeescript
PopoverManager.showPopover(
  target: 'target'
  content: PopoverManager.getContent 'content'
)
```

themis_components/thPopover/thPopover.directive.spec.coffee
```diff
@@ -1,49 +1,25 @@
...
+    it 'should throw an error', ->
+      expect(-> compileDirective("<div th-popover></div>")).toThrow()
 
-  it 'should close after it opens', ->
-    element.triggerHandler 'click'
-    angular.element(document.querySelector('.th-popover-overlay')).triggerHandler 'click'
+  describe 'when template is valid', ->
+    beforeEach ->
+      spyOn(PopoverManager, "attachPopover")
+      spyOn(PopoverManager, "getContent")
```
seanhealy: Do we need this spy?

themis_components/thPopover/thAddPopover.helper.coffee
dtomalty: Nope, that's literally all. I just did a diff to verify.

themis_components/thPopover/thPopover.directive.spec.coffee
dtomalty: I added the spy because the directive actually calls `PopoverManager.getContent` and it throws an error if the content isn't defined. I could easily define the content block in the html to get around this but I also liked the spec not being dependent on `PopoverManager` and `th-popover-content` which is tested elsewhere. Is there another way of doing this without adding the spy?

themis_components/thPopover/examples/4 - Trigger from script/coffee.coffee
dtomalty: I'm not sure. I like the dictionary syntax but I feel like having the params explicit in the definition makes it clearer that they're required. Just my thought, but I'm open to whatever convention we prefer.

themis_components/thPopover/thPopover.directive.coffee
dtomalty: Yah I like that.

themis_components/thPopover/examples/4 - Trigger from script/coffee.coffee
seanhealy: :white_check_mark: Resolved

themis_components/thPopover/examples/3 - Inline content/html.html
seanhealy: :white_check_mark: Resolved

themis_components/thPopover/readme.md
seanhealy: :white_check_mark: Resolved

themis_components/thPopover/thPopoverManager.service.coffee
seanhealy: :white_check_mark: Resolved

themis_components/thPopover/thPopover.directive.coffee
seanhealy: One question here. If we take this approach can you still reference things in the parent scope like you can with nested controllers?

```
- aController
  - popover
    - aController.foo
```

themis_components/thPopover/thAddPopover.helper.coffee
seanhealy: :white_check_mark: Resolved

themis_components/thPopover/thPopover.directive.spec.coffee
```diff
@@ -1,49 +1,25 @@
...
+  describe 'when name is not specified', ->
+    beforeEach ->
+      spyOn(PopoverManager, "attachPopover")
+      spyOn(PopoverManager, "getContent")
 
-  it 'should open overlay under the popover', ->
-    element.triggerHandler 'click'
-    expect(document.querySelector('.th-popover-overlay')).not.toBeNull()
+    it 'should throw an error', ->
+      expect(-> compileDirective("<div th-popover></div>")).toThrow()
```
seanhealy: :white_check_mark: Resolved

themis_components/thPopover/examples/4 - Trigger from script/coffee.coffee
seanhealy: Lets chat over this one when you have time. I see your point but also like the verboseness…

themis_components/thPopover/thPopover.directive.spec.coffee
seanhealy: Ahhh, so in this case you're using the spy like a mock?

seanhealy: :white_check_mark: Resolved

themis_components/thPopover/thAddPopover.helper.coffee
```diff
@@ -0,0 +1,124 @@
...
+
+    body = angular.element(document.body)
+    body.append overlay
+    body.append view
+    body.append arrow
+
+    overlay.on 'click', ->
+      dismissPopover()
+
+    $compile(view)(scope)
```
dtomalty: @seanhealy Note that the popover template is getting compiled against the scope that is passed in from the popover/url/target directive and not the scope of the content directive. This is because the scope isn't available when the content gets added in the compile phase but should this be ok in most cases.

themis_components/thPopover/thPopoverManager.service.coffee
```diff
@@ -0,0 +1,59 @@
...
+    showPopover = (options = {}) ->
+      {
+        targetName      # String: Required
+        contentPromise  # Promise: Required
+      } = options
+
+      unless targets.hasOwnProperty(targetName)
+        throw new Error "PopoverManager: target '#{targetName}' does not exist."
+
+      unless contentPromise.constructor.name is 'Promise'
```
dtomalty: @seanhealy is this a valid thing to do? Check the constructor name to see that its type is 'promise'

seanhealy: `instanceof` or `typeof` would be more normal with `instanceof` being the preferable of the two. That being said what you have there will work but if we can switch to `instanceof` it would be better.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof

themis_components/thPopover/readme.md
```diff
@@ -1,12 +1,18 @@
...
 
 Add small overlays of content, like those on the iPad, to any element for housing secondary information.
 
-`th-popover` accepts a URL of where it will fetch its inner content from.
+`th-popover` accepts the name of a `th-popover-content` element where its content will come from.
 
-`th-popover` also accepts an optional `overflow` param that will apply a styling class. This is intended to be used when dealing with legacy components that have elements that are positioned absolutely (such as datepickers). The overflow property is set to `auto` by default. Valid options:
+`th-popover-url` accepts a URL from which it will fetch its content.
+
+`th-popover-content` specifies a block of content used for display in a popover. The element must specify a unique `name` used to identify the content in `th-popover` and `PopoverManager` api.
```
seanhealy: 1) Long lines here.

2) If we're stuck with content being compiled within the scope of the target we should mention that here.

themis_components/thPopover/thAddPopover.helper.coffee
seanhealy: I have an idea here around this. Chat when you have a moment?

themis_components/thPopover/thPopoverManager.service.coffee
dtomalty: use presence of .then() instead

themis_components/thPopover/thPopover.directive.coffee
seanhealy: :white_check_mark: Resolved

themis_components/thPopover/readme.md
seanhealy: :white_check_mark: Resolved

themis_components/thPopover/examples/4 - Trigger from script/coffee.coffee
seanhealy: :white_check_mark: Resolved

themis_components/thPopover/thAddPopover.helper.coffee
seanhealy: :white_check_mark: Resolved

themis_components/thPopover/thPopoverManager.service.coffee
```diff
@@ -0,0 +1,63 @@
...
+    showPopover = (options = {}) ->
+      {
+        targetName        # String: Required
+        contentCallback   # Function: Required
+      } = options
+
+      unless targets.hasOwnProperty(targetName)
+        throw new Error "PopoverManager: target '#{targetName}' does not exist."
+
+      # unless getContentPromise instanceof Function
```
seanhealy: Is there a reason this is commented out?

themis_components/thPopover/thPopoverManager.service.coffee
seanhealy: :white_check_mark: Resolved

themis_components/thPopover/readme.md
seanhealy: :white_check_mark: Resolved

PR #102, opened 2016-03-04
clio-17509 Update to thAlert to fix possible message overflow problems.

JefCurtis:


themis_components/thInput/index.scss
```diff
@@ -1,10 +1,6 @@
 .th-input {
   outline: none;
-  -webkit-box-flex: 1;      // OLD - iOS 6-, Safari 3.1-6
-  -moz-box-flex: 1;         // OLD - Firefox 19-
-  -webkit-flex: 1;          // Chrome
-  -ms-flex: 1;              // IE 10
-  flex: 1;                  // NEW, Spec - Opera 12.1, Firefox 20+
+  flex: 1;
```
JefCurtis: We are using `gulp-autoprefixer` to take care of this for us.

themis_components/thAlert/index.scss
```diff
@@ -1,19 +1,26 @@
...
   left: 0;
-  width: calc(100% - 32px);
+  width: calc(100% - (#{$padding} * 2));
   color: #fff;
   z-index: 2147483647;
-  height: 54px;
-  padding-left: 16px;
-  padding-right: 16px;
-  display: block;
+  min-height: $height;
```
dtomalty: So previously it wouldn't grow with an overflowing message?

themis_components/thAlert/index.scss
```diff
@@ -1,19 +1,26 @@
...
+  padding-left: $padding;
+  padding-right: $padding;
+  display: flex;
   @include animated(300ms);
 
   &.ng-animate {
     &.ng-hide {
-      top: -54px;
+      top: -#{$height};
+      opacity: 0;
```
dtomalty: What hid it before?

themis_components/thAlert/index.scss
```diff
@@ -53,10 +60,11 @@
     &.alert-icon {
       color: #fff;
       text-align: center;
-      width: 40px;
-      height: 40px;
-      line-height: 40px;
-      margin-right: 40px;
+      min-width: $alert-icon-size;
```
dtomalty: Why is this `min-width` now?

themis_components/thAlert/index.scss
```diff
@@ -38,8 +45,8 @@
   }
 
   p {
-    display: inline-block;
-    width: calc(100% - 104px);
```
dtomalty: This is handled by the flexbox now?

themis_components/thInput/index.scss
JefCurtis: :white_check_mark: Resolved

themis_components/thAlert/index.scss
JefCurtis: Yes, before it would always be a height of `54px`.

themis_components/thAlert/index.scss
JefCurtis: It's animating upward from `top: 54px`  to `top: -54px` but because there is a chance that the height of the banner can now be bigger than `54px` we are setting the opacity to 0 to hide any subtle flicker that might occur. This is really not a common situation but adding the `opacity: 0` will make it basically imperceivable.

themis_components/thAlert/index.scss
```diff
@@ -53,10 +60,11 @@
...
       text-align: center;
-      width: 40px;
-      height: 40px;
-      line-height: 40px;
-      margin-right: 40px;
+      min-width: $alert-icon-size;
+      height: $alert-icon-size;
+      line-height: $alert-icon-size;
+      margin-right: $margin;
+      margin-top: ($height - $margin) / 2;
```
JefCurtis: So the `margin-top` is going to be `7px` but now it's calculated instead of hard coded based on the size of the icon. I just wanted to remove any magic numbers if I could.

themis_components/thAlert/index.scss
JefCurtis: Not necessary. Changing this back.

themis_components/thAlert/index.scss
JefCurtis: Yes the width of the `p` element changes on it's own so no need for the `calc` call.

themis_components/thAlert/index.scss
```diff
@@ -66,8 +74,10 @@
 
     &.dismiss-icon {
       cursor: pointer;
-      line-height: 54px;
-      width: 16px;
+      line-height: $height;
+      min-width: $dismiss-icon-size;
+      margin-left: $margin;
```
JefCurtis: When the `p` text grows very long it pushes up against the dismiss icon. It should have been there before this fix but it was overlooked.

themis_components/thAlert/index.scss
dtomalty: Resolved

dtomalty: Resolved

themis_components/thAlert/index.scss
dtomalty: Resolved

themis_components/thAlert/index.scss
dtomalty: Resolved

themis_components/thAlert/index.scss
dtomalty: Resolved

themis_components/thAlert/index.scss
dtomalty: Resolved

themis_components/thAlert/index.scss
dtomalty: Resolved

PR #119, opened 2016-04-11
CLIO-33515: Add truncation to thSelect

craigcarlyle:


themis_components/thSelect/index.scss
```diff
@@ -16,7 +16,7 @@
     box-sizing: border-box;
 
     margin: 0;
-    padding: 12px;
+    padding: 12px 24px 12px 12px;
```
JefCurtis: I was just looking that this and its seems like the text is pretty tight up against the caret icon. What do you think about making the `padding-right: 36px` instead of `24px`?

Before
padding-right 24px

After
padding-right 36px

PR #131, opened 2016-04-27
Clio 15749 filter clear

dtomalty:


themis_components/thFilter/filters/input/thFilter.input.directive.coffee
```diff
@@ -22,4 +22,7 @@ angular.module 'ThemisComponents'
         if @filter.getValue()?
           @filterSet.onFilterChange()
 
+      $scope.$on "clearFilters", =>
```
seanhealy: It's usually a good idea to namespace events. Perhaps we should call this something like: `th-filters-clear`?

PR #132, opened 2016-04-29
CLIO-35426: Add icon support to thAutocomplete.

dtomalty:


themis_components/thAutocomplete/index.scss
```diff
@@ -100,11 +103,11 @@ th-autocomplete {
     }
 
     .caret {
-      display: inline-block;
+      display: none;
       width: 0;
```
JeffMarvin: These other properties can be ommitted if this is not being displayed.

themis_components/thAutocomplete/index.scss
```diff
@@ -1,6 +1,9 @@
 th-autocomplete {
   @import "../../node_modules/ui-select/dist/select";
 
+  display: block;
+  position: relative;
+
```
JeffMarvin: Are these here so you can do the absolute positioning on .th-autocomplete-icon? (Just wondering.)

themis_components/thAutocomplete/index.scss
dtomalty: Good call :)

themis_components/thAutocomplete/index.scss
dtomalty: Exactly. The positioning of the child icon element doesn't work without this.

themis_components/thAutocomplete/index.scss
```diff
@@ -100,17 +103,7 @@ th-autocomplete {
...
-      height: 0;
-      margin-left: 2px;
-      
-      vertical-align: middle;
-
-      border-top: 4px dashed;
-      border-top: 4px solid \9;
-      border-right: 4px solid transparent;
-      border-left: 4px solid transparent;
+      display: none;
```
seanhealy: If an icon isn't specified do we still get / want the `^`?

dtomalty: Spoke with design, they don't want the `^`. Logic was `^` gives the impression that it's something you can click and expect to see a drop down, which is not the case with autocomplete. The drop down reveals itself when you start typing, so doesn't need the indicator.

themis_components/thAutocomplete/index.scss
JeffMarvin: :white_check_mark: Resolved

themis_components/thAutocomplete/index.scss
JeffMarvin: :white_check_mark: Resolved

themis_components/thAutocomplete/index.scss
JeffMarvin: :white_check_mark: Resolved

PR #138, opened 2016-05-13
CLIO-35366: Set min-width attribute on input.

dtomalty:


themis_components/thInput/index.scss
```diff
@@ -10,6 +10,7 @@
   outline: none;
   flex: 1;
   width: 100%;
+  min-width: 60px;
```
dtomalty: Established with @gthierman. Accommodates three characters, rounded up, including padding.

PR #146, opened 2016-05-26
CLIO-34343: Added sub-header styling to thTabset

craigcarlyle:


themis_components/thTabset/index.scss
```diff
@@ -22,4 +22,17 @@
       }
     }
   }
-}
\ No newline at end of file
+
+  &.sub-header {
```
PANCAKES: I think ```sub-header``` feels wrong. It makes sense in the specific case e are using it here, but in the more general sense, it is just larger? Talked to @seanhealy and maybe ```kind="large"``` is more accurate. That way, we can add other kinds later, if needed.

themis_components/thTabset/readme.md
```diff
@@ -19,6 +19,9 @@ using `ng-click`. Using `ng-click` will not effect the normal usage of `thTabset
 but just allow a click event to attach to. See example 3 below.
 [Angular docs](https://docs.angularjs.org/api/ng/directive/ngClick)
 
+[sub-header="boolean"]
+- `sub-header` denotes wether the tabset is a sub-header or not. As a sub-header, the `thTabset` styling matches that used with the `thContentHeader` component.
```
PANCAKES: Have the docs focus on the size difference and that it matches up with ```thContentHeader``` in that respect, perhaps.

themis_components/thTabset/examples/4 - Alternate Style/html.html
```diff
@@ -0,0 +1,16 @@
+<html ng-app="thDemo">
+  <body>
+
+    <th-tabset sub-header="true">
```
PANCAKES: ```<th-tabset kind="large">```

themis_components/thTabset/thTabset.directive.coffee
```diff
@@ -4,7 +4,8 @@ angular.module('ThemisComponents')
     replace: true
     template: require './thTabset.native.template.html'
     transclude: true
-    scope: {}
+    scope:
+      type: "@"
```
seanhealy: We should check the type that is passed in against a list of acceptable types. We want to enforce that people aren't inventing them on the fly.

PR #149, opened 2016-06-01
clio-34452 add [th-popover-persist] to whitelist

seanhealy:


themis_components/thPopover/readme.md
```diff
@@ -26,8 +26,10 @@ datepickers). The overflow property is set to `auto` by default. Valid options:
   - `visible`
   - `hidden`
 
-## Usage
+Adding the attribute `th-popover-persist` to any element within the popover will
```
dtomalty: "to any element" ..I think it only applies to anchor elements? (`...and event.target.matches("a, a *")`)

PR #154, opened 2016-06-15
Clio 37534 thfilter email url

dtomalty:


themis_components/thFilter/filters/input/thFilter.input.directive.coffee
```diff
@@ -5,6 +5,7 @@ angular.module 'ThemisComponents'
       filterSet: '='
```
dtomalty: Changes in these 4 files are so I can pass in `placeholder` directly from the element, but `placeholder` coming from the filter options still takes precedence.

themis_components/thFilter/thCustomFilterRow.template.html
```diff
@@ -28,6 +28,22 @@
       ng-blur="thCustomFilters.filterSet.onFilterChange()"
       >
     </th-filter-input>
+    <th-filter-input
+      ng-switch-when="email"
```
JefCurtis: My only suggestion is that we should probably be passing the input type along to the underlying `th-input` component. 
ex: `email` and `url

PR #156, opened 2016-06-15
clio-37675: Add the ability to set an active tab programmatically

JefCurtis:


PANCAKES:
Should we add anything to the docs for thTabSet to let the developer know they can use this new feature?

PR #158, opened 2016-06-16
CLIO-37529: thFilter checkbox custom filter field.

dtomalty:
…tive. Use select directive for checkbox.


themis_components/thFilter/thCustomFilterRow.template.html
```diff
@@ -28,6 +28,13 @@
       ng-blur="thCustomFilters.filterSet.onFilterChange()"
       >
     </th-filter-input>
+    <th-filter-select
```
JefCurtis: Any reason we can't use a `th-checkbox` or even better `th-switch` here instead of the select?

themis_components/thFilter/filters/select/thFilter.select.directive.coffee
```diff
@@ -4,6 +4,7 @@ angular.module 'ThemisComponents'
     scope:
       filterSet: '='
       filterOptions: '='
+      options: '='
```
JefCurtis: Any changes needed to the readme here?

dtomalty: No as these directives are only used internal to the component. The interface to the user-facing directives hasn't changed.

themis_components/thFilter/thCustomFilterRow.template.html
dtomalty: No reason we can't. For now this matches filters in themis.

PR #159, opened 2016-06-20
CLIO-37535: thFilter number custom field type.

dtomalty:


themis_components/thInput/index.scss
```diff
@@ -1,9 +1,11 @@
 .th-input-wrapper {
```
dtomalty: The changes to this file effectively move the height to the outer wrapper element, which is where the border is specified, and so we can effectively set border-box sizing and have it size properly. Previously, the height was specified on the inner `.th-input` element, but the border was specified on the wrapper, so it would size incorrectly.

themis_components/thTextarea/thTextarea.directive.spec.coffee
```diff
@@ -4,9 +4,11 @@
 context = describe
```
dtomalty: Was getting an error in this test, suspect it has to do with asynchronous-terminating `done` call.

themis_components/thFilter/filters/number/numberFilter.service.coffee
```diff
@@ -0,0 +1,28 @@
+angular.module "ThemisComponents"
+.factory "NumberFilter", (FilterBase) ->
+  class NumberFilter extends FilterBase
+    constructor: (options = {}, operatorOptions  = [], defaultOperatorIndex = 0) ->
+      super options
+      @placeholder = options.placeholder
```
JeffMarvin: Should an error be thrown if options.placeholder doesn't exist?

themis_components/thFilter/filters/number/thFilter.number.template.html
```diff
@@ -0,0 +1,19 @@
...
+    ng-model="thFilterNumber.filter.operator"
+    ng-change="thFilterNumber.onOperatorChange()"
+    options="thFilterNumber.filter.operatorOptions"
+    condensed="true"
+    class="operator"
+    >
+  </th-select>
+  <th-input
+    ng-model="thFilterNumber.filter.model"
+    ng-attr-placeholder="{{thFilterNumber.filter.placeholder || 'Type a number…'}}"
```
JeffMarvin: We should consider localizing this string.

themis_components/thFilter/filters/number/numberFilter.service.coffee
dtomalty: If it doesn't exist, `@placeholder` will be undefined, which is handled in the template with the `||`.

themis_components/thFilter/filters/number/thFilter.number.template.html
dtomalty: Agreed. But this is consistent with the other filters for now. It'll need a whole localization pass..

themis_components/thFilter/filters/number/numberFilter.service.coffee
JeffMarvin: Maybe that logic is better here than in the view? I'm open to disagreements. ;)

themis_components/thFilter/filters/number/thFilter.number.template.html
JeffMarvin: Ok. Resolved.

themis_components/thFilter/filters/number/numberFilter.service.coffee
dtomalty: In this case I prefer it in the view because there's a pretty clear separation between the model and view here. I don't think the `NumberFilter` model should care about what the default text in the view is.

PR #162, opened 2016-06-23
Clio 37536 thfilter money

dtomalty:


public/json/customFields.json
```diff
@@ -15,6 +15,11 @@
     "id": "4"
   },
   {
+    "name": "money",
```
JefCurtis: The description in the `CustomField` serializer is "currency". Should this match?

themis_components/thFilter/thCustomFilterRow.directive.coffee
```diff
@@ -21,6 +21,11 @@ angular.module 'ThemisComponents'
         {name: ">=", value: ">="}
         {name: ">", value: ">"}
       ]
+      @moneyOperatorOptions = [
+        {name: "Less than", value: "<"}
```
JefCurtis: What about `>=` or `<=`? Those are both available as filters in the API.

dtomalty: This matches the existing money filter. If we want to change it it'd be as easy as adding to this array.

public/json/customFields.json
dtomalty: Yep good catch- updated.

PR #167, opened 2016-07-06
clio-38286: Upgrade Angular to 1.5.7

JefCurtis:
- Update Angular from 1.4.7 to 1.5.7
- Update ui-select from 0.13.1 to 0.18.0
- Fixed for ui breaking changes in ui-select
- Fix for several directives where isolated scoped variable should be
optional


themis_components/thAutocomplete/index.scss
```diff
@@ -179,11 +179,25 @@ th-autocomplete {
...
-    }
-
-    input {
       font-size: inherit;
       background: $off-white;
+
+      min-width: 100%;
+
+      padding: 12px;
+      padding-right: 40px;
```
dtomalty: Would prefer these on one line, so one isn't missed if they need to be changed.

dtomalty: Also, right padding used to take into account the icon. Does it still?

themis_components/thAutocomplete/index.scss
```diff
@@ -179,11 +179,25 @@ th-autocomplete {
 
     .ui-select-search {
       font-family: inherit;
-    }
-
-    input {
       font-size: inherit;
       background: $off-white;
+
+      min-width: 100%;
```
dtomalty: What is this used for?

themis_components/thFilter/filters/number/thFilter.number.directive.spec.coffee
```diff
@@ -85,7 +85,7 @@ describe "ThemisComponents: Directive: thFilterNumber", ->
         setInputValue "1000"
 
       it "should call onFilterChange", ->
-        selectOperator operatorOptions[1].value
+        selectOperator operatorOptions[1]
```
dtomalty: There's another instance in the previous test..

themis_components/thAutocomplete/index.scss
JefCurtis: It does accommodate an icon but left the padding in regardless if there was an icon or not. Updated to add padding conditionally.

themis_components/thAutocomplete/index.scss
JefCurtis: This was causing crazy resizing with the window observer. By setting it to have a min width of 100% there was no need for it to resize.

themis_components/thAutocomplete/index.scss
dtomalty: :white_check_mark: Resolved

themis_components/thAutocomplete/index.scss
dtomalty: :white_check_mark: Resolved

themis_components/thFilter/filters/number/thFilter.number.directive.spec.coffee
dtomalty: :white_check_mark: Resolved

PR #168, opened 2016-07-12
clio-38294 address ws vulnerability

JeffMarvin:


package.json
```diff
@@ -20,7 +20,7 @@
   "dependencies": {
     "angular": "1.4.7",
     "angular-animate": "1.4.7",
-    "angular-datepicker": "g00fy-/angular-datepicker#9842705e6dc90fd265e1a196ee444142ad99103c",
+    "angular-datepicker": "git://github.com/g00fy-/angular-datepicker.git#9842705e6dc90fd265e1a196ee444142ad99103c",
```
JeffMarvin: Just adjusting the way these are linked. No version change.

package.json
```diff
@@ -44,13 +43,14 @@
     "node-bourbon": "^4.2.1-beta1",
     "qs": "^6.1.0",
     "stringify": "^3.1.0",
-    "ui-select": "angular-ui/ui-select.git#v0.13.1",
+    "ui-select": "git://github.com/angular-ui/ui-select.git#v0.13.1",
```
JeffMarvin: Just adjusting the way these are linked. No version change.

gulp/tasks/docsServer.coffee
```diff
@@ -5,7 +5,6 @@ fs = require 'fs'
 coffeescript = require 'coffee-script'
 
 express = require 'express'
-expressWs = require 'express-ws'
```
JeffMarvin: This and related expressWs code moved to within the docs-server task, so that they are not required in dependencies, only in devDepedencies.

package.json
```diff
@@ -30,7 +30,6 @@
     "coffeeify": "^2.0.1",
     "es6-promise": "~3.0.2",
     "express": "^4.12.4",
-    "express-ws": "^0.2.6",
```
JeffMarvin: The real security solution. Moving `express-ws`, which uses `ws`, into a devDependency. This means that Themis won't need `ws` in it's node_modules folder, hence addressing the vulnerability.

PR #173, opened 2016-08-08
CLIO-39365: First pass of autocomplete multiselect.

dtomalty:


themis_components/thAutocomplete/thAutocomplete.directive.coffee
```diff
@@ -62,3 +72,13 @@ angular.module 'ThemisComponents'
 
       compiledTemplate = $compile(templateElement) childScope
       element.append compiledTemplate
+
+      if multiple
+        $timeout ->
```
dtomalty: This is on a `$timeout` because the current digest cycle has to complete for the `ui-select` to finish initializing. `.ui-select-search` and `.ui-select-container` don't exist during linking. 

This is taking the focus from the input and applying it to the container for highlighting.

themis_components/thAutocomplete/thAutocomplete.directive.coffee
```diff
@@ -1,5 +1,5 @@
 angular.module 'ThemisComponents'
-  .directive 'thAutocomplete', ($compile, $interpolate) ->
+  .directive 'thAutocomplete', ($compile, $interpolate, $timeout) ->
     restrict: 'E'
     scope:
       ngModel: '=?'
```
frankieyan: This and `multiple` should probably be marked as optional in [readme.md](https://github.com/dtomalty/lib-themisui/blob/4f1de0a2348a82ac2b8637894af42b11920c19c3/themis_components/thAutocomplete/readme.md)

themis_components/thAutocomplete/thAutocomplete.directive.coffee
frankieyan: Is it possible to throw this in the postLink function?

themis_components/thAutocomplete/index.scss
```diff
@@ -5,9 +5,142 @@
...
+
+    font: inherit;
+
+    background: $off-white;
+    border: 1px solid $grey-70;
+    border-radius: 4px;
+
+    &:focus,
+    &.has-focus {
+      box-shadow: 0 0 5px rgba(97, 179, 227, .7);
```
frankieyan: Hex or pull in a variable?

themis_components/thAutocomplete/thAutocomplete.directive.coffee
```diff
@@ -62,3 +72,13 @@ angular.module 'ThemisComponents'
...
+
+      if multiple
+        $timeout ->
+          # Toggle container shadow when input has focus.
+          search = element.querySelectorAll(".ui-select-search")
+          container = element.querySelectorAll(".ui-select-container")
+          search.on "focus", ->
+            container.addClass("has-focus")
+          search.on "blur", ->
+            container.removeClass("has-focus")
```
frankieyan: I was wondering how this was possible since `querySelectorAll` is a method on native elements and `element` was a jqLite element - traced it back to ui-select which had this snippet:
```
/**
 * Add querySelectorAll() to jqLite.
 *
 * jqLite find() is limited to lookups by tag name.
 * TODO This will change with future versions of AngularJS, to be removed when this happens
 *
 * See jqLite.find - why not use querySelectorAll? https://github.com/angular/angular.js/issues/3586
 * See feat(jqLite): use querySelectorAll instead of getElementsByTagName in jqLite.find https://github.com/angular/angular.js/pull/3598
 */
```

I _feel_ like we should try to avoid using this - especially when it's not documented in the jqLite documentations, and instead explicitly do what this method does behind the scenes: 
```
angular.element(this[0].querySelectorAll(selector));
```

themis_components/thAutocomplete/thAutocomplete.directive.coffee
dtomalty: My understanding is that this `link` function is the same as having it in `postLink` of `compile`. I'm speculating `ui-select` does something on a `timeout`, as without the `timeout` the search and container elements aren't yet available.

themis_components/thAutocomplete/thAutocomplete.directive.coffee
dtomalty: Fixed this.

themis_components/thAutocomplete/thAutocomplete.directive.coffee
dtomalty: Updated.

themis_components/thAutocomplete/thAutocomplete.directive.coffee
frankieyan: :white_check_mark: Resolved

themis_components/thAutocomplete/thAutocomplete.directive.coffee
frankieyan: :white_check_mark: Resolved

themis_components/thAutocomplete/index.scss
frankieyan: :white_check_mark: Resolved

themis_components/thAutocomplete/thAutocomplete.directive.coffee
frankieyan: :white_check_mark: Resolved

PR #175, opened 2016-08-08
Add version to docs viewer.

seanhealy:


public/stylesheets/docs-app.scss
```diff
@@ -32,10 +32,30 @@ body {
     background-color: $blue;
 
     h1 {
+      display: block;
+      position: relative;
+      width: 62px;
+
+      margin-left: auto;
```
JefCurtis: would `margin: auto;` do basically the same thing?

public/javascript/controllers/AppController.coffee
```diff
@@ -1,4 +1,12 @@
 angular = require 'angular'
 
 angular.module('ThemisComponentsApp')
-  .controller 'AppController', -> return
+  .controller 'AppController', ($scope, $http) ->
+
+    $scope.version = "…"
+
+    $http.get "/version.json"
```
JefCurtis: This looks weird to me as it's not indented here. Something like:
```coffeescript
$http:get("/version.json").then (response) ->
  $scope.version = "v#{response.data.version}"
```

public/javascript/controllers/AppController.coffee
```diff
@@ -1,4 +1,12 @@
 angular = require 'angular'
 
 angular.module('ThemisComponentsApp')
-  .controller 'AppController', -> return
+  .controller 'AppController', ($scope, $http) ->
+
+    $scope.version = "…"
```
JefCurtis: Should it just be blank if there is no version number? Of course I can't see how this could happen but it would look weird to just have an ellipsis climbing up the right side of the sidebar header.

seanhealy: Like a little tiny worm!

public/stylesheets/docs-app.scss
```diff
@@ -36,6 +36,21 @@ body {
...
+
+    .version {
+      display: block;
+      position: absolute;
+      top: 15px;
+      right: -10px;
+      width: $header-height;
+
+      line-height: normal;
+      font-size: 8.5px;
```
seanhealy: That really depends on browser / OS.

Fonts in general are REALLY inconsistent… 😢

public/stylesheets/docs-app.scss
seanhealy: ish… I didn't want to mess with the top / bottom margin though.

PR #184, opened 2016-08-23
CLIO-36305: Add required indicator to thWithLabel

craigcarlyle:
* Modified `thWithLabel` to include a _"required"_ indicator if the component is using `ng-required`
* Updated `thWithLabel` spec to reflect the change


themis_components/thWithLabel/index.scss
```diff
@@ -21,6 +21,10 @@ label.th-label {
     color: $light-grey;
     font-weight: bold;
     margin-bottom: 8px;
+  }
 
+  .required-field {
+    float: right;
+    color: #55a51c;
```
JefCurtis: Is there a variable color for this?

themis_components/thWithLabel/examples/4 - Required/html.html
```diff
@@ -0,0 +1,65 @@
+<html ng-app="thDemo">
```
JefCurtis: Do we need another example for this? I feel like we could just add a few `ng-require`s to some of the other examples.

themis_components/thWithLabel/index.scss
```diff
@@ -21,6 +21,10 @@ label.th-label {
     color: $light-grey;
     font-weight: bold;
     margin-bottom: 8px;
+  }
 
+  .required-field {
+    float: right;
```
seanhealy: We should avoid using float for positioning.

PR #194, opened 2016-09-06
clio-36232 fix datepicker popup bug in Themis

JeffMarvin:


themis_components/thPopover/thAddPopover.helper.coffee
```diff
@@ -123,6 +123,8 @@ module.exports = ($compile, $timeout) -> (target, contentCallback) ->
           '.th-switch *'
           '[th-popover-persist]'
           '.ui-select-choices-row-inner *' # Select 2 autocompleter
+          '.ui-datepicker-prev'
```
frankieyan: Would increase the scope of this quite a bit - but it'd save us maintenance time down the road if we don't have to keep the whitelist/blacklist in `th-popover` updated so it can remain ignorant of its usage. Instead, the only default whitelist/black list could simply be `[th-popover-persist]` and `a *`. 

The `PopoverManager` then could provide a method i.e. `PopoverManager.addWhiteList` or `PopoverManager.addBlackList` to add additional selectors to the service when initialized in the consuming controller. Similarly, in use cases where the attribute directives are being used i.e. `[th-popover-url]`, we can provide an attribute on the `$element` that takes in an array and pass the provided list to the manager.

PR #198, opened 2016-09-09
CLIO-40306: Fix for misusing valueField; it should only be used for o…

dtomalty:
…ptions url path.


themis_components/thFilter/filters/select/selectFilter.service.coffee
```diff
@@ -10,8 +10,9 @@ angular.module "ThemisComponents"
       @valueField = options.selectOptionsValueField or "value"
 
       if options.selectOptionsUrl?
-        @model = {}
-        @model[@valueField] = initialState?.value
+        @model = {
```
dtomalty: The gist of this fix, here and below is that `@valueField` should only be used in the url path (see line 26). Once they are mapped there, we should always be using `.value`.

JefCurtis: Little more succinct? 
```coffeescript
@model = value: initialState?.value
```

PR #207, opened 2016-09-13
CLIO-40324: thDropdown improvements

craigcarlyle:
* Allowed `ng-disabled` attribute to be set for `thDisclosure`
* Added the `text-side` attribute to `thDisclosure` to customize which side of the caret the text is placed
* Added two examples for each of the new attributes and updated the `thDisclosure` readme


themis_components/thDisclosure/examples/5 - ngDisabled/html.html
```diff
@@ -0,0 +1,19 @@
+<html ng-app="thDemo">
+  <body
+    style="height: 25px;"
+    ng-controller="DemoCtrl as demo"
+    >
+    <th-disclosure-toggle
+      name="unique-name"
+      ng-disabled="true"
```
PANCAKES: Is this currently just a static attribute? I think the developer will want the ability to enable this after the page is rendered. In the example of Shaheer's work, they will complete steps 1,2 then enable the 3rd step (the disabled disclosure) in code.

craigcarlyle: It's static in this particular example but it can be set to a variable and changed on the fly.

PR #211, opened 2016-09-15
Clio 39851 tabset badges

frankieyan:


themis_components/thTabset/readme.md
```diff
@@ -19,6 +20,10 @@ using `ng-click`. Using `ng-click` will not effect the normal usage of `thTabset
 but just allow a click event to attach to. See example 3 below.
 [Angular docs](https://docs.angularjs.org/api/ng/directive/ngClick)
 
+[badge="number"]
+- Optional: `badge` allows a number value to be displayed as a badge beside the tab. See example 7 below.
```
JeffMarvin: Example 6, actually. ;)

themis_components/thTabset/thTabset.directive.coffee
```diff
@@ -32,6 +32,9 @@ angular.module('ThemisComponents')
       $scope.setActiveTab = ->
         $scope.activateTab tab for tab in tabs when tab.name is $scope.activeTab
 
+      $scope.hasBadge = (tab) ->
+        typeof tab.badge is "number"
```
seanhealy: Wonder if we should just check for existence. I could be a valid case to have a non number in a badge perhaps such as `99+`, `new`.

seanhealy:
That red box is angrier than I expected / wanted it to be. :p

themis_components/thTabset/index.scss
```diff
@@ -61,4 +66,20 @@
...
     }
   }
+
+  .th-tab-badge {
+    display: inline-block;
+    margin-left: 10px;
+    padding: 3px 8px;
+
+    color: $grey;
+    font-size: 0.7896em;
```
frankieyan: It's the datatable badges' font size / default th-tab font size. Trying to get it as close to the datatable as possible here so I have a bunch of hardcoded px values too....

themis_components/thTabset/readme.md
frankieyan: Actually it is 7... I've made the bigger mistake of naming the folder wrong 😄

PR #213, opened 2016-09-18
clio-40556 datepicker custom filter operator

JeffMarvin:


themis_components/thFilter/filters/date/dateFilter.service.coffee
```diff
@@ -3,18 +3,38 @@ moment = require "moment"
...
+      operatorOptions = []
+      defaultOperatorIndex = 0
+      initialState
+    ) ->
       super options
       @placeholder = options.placeholder
 
+      # define operator if this is a custom filter
+      @operatorOptions ?= operatorOptions
+      @defaultOperatorIndex ?= defaultOperatorIndex
```
seanhealy: Rather than doing this if I'm understanding what you're wanting here properly you can also do the following above.

```coffeescript
   class DateFilter extends FilterBase
    constructor: (
      options = {}
      @operatorOptions = []
      @defaultOperatorIndex = 0
      initialState
    ) ->
```

themis_components/thFilter/filters/date/dateFilter.service.coffee
```diff
@@ -3,18 +3,38 @@ moment = require "moment"
...
+
+      @clearState()
+
+      if initialState
```
seanhealy: `if initialState?`

themis_components/thFilter/filters/date/dateFilter.service.coffee
JeffMarvin: Oh, handy. Fixed.

themis_components/thFilter/filters/date/dateFilter.service.coffee
JeffMarvin: yes, better.

themis_components/thFilter/filters/date/dateFilter.service.coffee
```diff
@@ -3,18 +3,34 @@ moment = require "moment"
...
+      @defaultOperatorIndex = 0
+      initialState
+    ) ->
       super options
       @placeholder = options.placeholder
 
+      @clearState()
+
+      if initialState?
+        @model = moment initialState.value.format
```
dtomalty: Not clear where `format` is coming from..I don't see it in the example above. Looks like this might throw an error?

themis_components/thFilter/filters/date/dateFilter.service.coffee
```diff
@@ -3,18 +3,34 @@ moment = require "moment"
...
+
+        if @operator?.value?
+          state.operator = @operator.value
+
+        return state
       else return null
 
     clearState: =>
       @model = null
+      @operator ?= @operatorOptions[@defaultOperatorIndex]
```
dtomalty: Wouldn't we want to set this value regardless? When the user hits clear filter I would expect it to go to the default operator..

themis_components/thFilter/filters/date/dateFilter.service.coffee
```diff
@@ -3,18 +3,34 @@ moment = require "moment"
...
     type: "date"
 
     getState: =>
       if @model?.isValid
-        return {
-          value: @model.format()
-        }
+        state = {value: @model.format()}
+
+        if @operator?.value?
```
dtomalty: Is this because the filter won't necessarily have an operator associated with it? Do you think the api would still be expecting one? If that's the case, maybe we need to pass a default operator as part of `getState`

themis_components/thFilter/filters/date/thFilter.date.template.html
```diff
@@ -1,9 +1,19 @@
-<div>
+<div ng-class="{inner: thFilterDate.hasOperator}">
```
dtomalty: Nice- interesting to see how you handled this! I'll probs have to apply something like this to the number/currentcy filters in the future..

themis_components/thFilter/filters/date/dateFilter.service.coffee
JeffMarvin: Based on some preliminary work on Themis Billable Clients integrating these date filters, the API was looking for an operator on the date custom filter, and not on the date static filter.

PR #214, opened 2016-09-19
Clio 37532 filter time

dtomalty:


themis_components/thTableActionBar/examples/2 - Remote/coffee.coffee
```diff
@@ -62,9 +62,11 @@ angular.module 'thDemo', ['ThemisComponents']
           _limit: pageSize
           _start: (currentPage - 1) * pageSize
           _end: ((currentPage - 1) * pageSize) + pageSize
-        searchString = @filterSet.getQueryString()
+        state = @filterSet.getState()
```
dtomalty: This fixes the action bar example as it was broken when the `FilterSet` api changed.

themis_components/thFilter/filters/time/timeFilter.service.coffee
```diff
@@ -0,0 +1,69 @@
+moment = require "moment"
+
+angular.module "ThemisComponents"
```
JeffMarvin: Need the new indent style here too.

themis_components/thFilter/filters/time/thFilter.time.directive.coffee
```diff
@@ -0,0 +1,55 @@
...
+    controller: ($scope, $element) ->
+      enterEventCode = 13
+      @operatorOptions = [
+        {name: "Before", value: "<"}
+        {name: "Exactly", value: "="}
+        {name: "After", value: ">"}
+      ]
+
+      setValid = ->
+        $timeout -> $element.querySelectorAll(".th-input-wrapper").removeClass("is-invalid")
```
JeffMarvin: This works neatly/is a tidy solution. :) Does design approve?

themis_components/thFilter/filters/time/timeFilter.service.coffee
```diff
@@ -0,0 +1,69 @@
...
+          item.value is initialState.operator
+        @validate()
+
+    type: "time"
+
+    validate: =>
+      if @model?.length > 0
+        @time = moment(@model,
+          [
+            "H:ma"
```
JeffMarvin: Does this match our parsing logic in the current Themis app for this field?

themis_components/thFilter/filters/time/timeFilter.service.coffee
```diff
@@ -0,0 +1,69 @@
...
+            "Ha"
+            "H a"
+            "H:m"
+            "Hm"
+            "H"
+          ]
+          true
+        )
+        if @time.isValid()
+          @model = @time.format("h:mm A")
```
JeffMarvin: Can drop the parentheses.

themis_components/thFilter/filters/time/thFilter.time.directive.coffee
```diff
@@ -0,0 +1,55 @@
...
+      filterSet: '='
+      filterOptions: '='
+      ngBlur: '&'
+      initialState: '=?'
+    bindToController: true
+    controllerAs: 'thFilterTime'
+    template: require './thFilter.time.template.html'
+    controller: ($scope, $element) ->
+      enterEventCode = 13
+      @operatorOptions = [
```
dtomalty: You're right, some of the other directives pass them in but that's because the number/currency filters share a directive but they have different operator options. In this case I don't anticipate the `Time` field having other options..

themis_components/thFilter/filters/time/thFilter.time.directive.coffee
```diff
@@ -0,0 +1,55 @@
+angular.module 'ThemisComponents'
```
dtomalty: Done.

themis_components/thFilter/filters/time/thFilter.time.directive.coffee
dtomalty: This is how it exists currently, so I didn't get explicit design signoff. It's unchanged.

themis_components/thFilter/filters/time/thFilter.time.directive.coffee
```diff
@@ -0,0 +1,55 @@
...
+        setValid()
+        @filter.clearState()
+
+      return
+    compile: ->
+      pre: (scope, element, attrs, controller) ->
+        controller.filter = new TimeFilter(
+          controller.filterOptions
+          controller.operatorOptions
+          1
```
dtomalty: Good call.

themis_components/thFilter/filters/time/thFilter.time.directive.coffee
dtomalty: I just spoke with Chris and the new style is for `.module` to be on the same line..updated to reflect this in both instances..

themis_components/thFilter/filters/time/timeFilter.service.coffee
dtomalty: Done.

themis_components/thFilter/filters/time/timeFilter.service.coffee
dtomalty: Yes it does, from javascripts/validation.js.erb:
```
var date = Date.parseExact(jq(this).val(), ["H:mtt", "Hmtt", "H:m tt", "Hm tt", "Htt", "H tt", "Hm", "H" ])
```
The only one I added was `"H:m"`...I think the reason this was excluded from the original list is because it's a default formatted that can be handled by the api. So if the string is in this format it doesn't need to be parsed, it just gets passed through directly.

themis_components/thFilter/filters/time/thFilter.time.directive.coffee
JeffMarvin: ok sounds good :)

PR #218, opened 2016-09-21
CLIO-40081: Replace thFilter select with an autocompleter

craigcarlyle:
* Replaced `thFilter`'s `customFilterTypes` select control with `thAutocomplete`
 * Initial filter type list is limited to 10 results
 * Filtering begins after 2+ characters have been entered
 * Added caret styling
* Reduced vertical padding of `thAutocomplete`'s list items when using the condensed styling


themis_components/thAutocomplete/index.scss
```diff
@@ -344,6 +344,12 @@
     .ui-select-container {
       height: 34px;
 
+      .ui-select-choices-row {
+        a {
+          padding: 5px 12px;
```
craigcarlyle: This reduces the vertical padding of the list items when `thAutocomplete` is using the condensed styling. The condensed styling reduces the input's vertical padding by `4px` in both directions, so the same measurement is applied here by reducing it from `9px` to `5px`.

themis_components/thFilter/index.scss
```diff
@@ -31,6 +31,20 @@
     margin-bottom: 0;
   }
 
+  .th-autocomplete-wrapper {
```
craigcarlyle: Applying these styles here instead of the `thAutocomplete` stylesheet as they're specific to `thFilter` and won't/shouldn't be used anywhere else.

themis_components/thFilter/index.scss
```diff
@@ -31,6 +31,20 @@
...
+    .ui-select-choices {
+      max-height: 292px;
```
craigcarlyle: Increasing the max-height here so the same number of choices are visible as our legacy custom filter control currently displays (10 items).

themis_components/thFilter/index.scss
```diff
@@ -31,6 +31,20 @@
...
+    }
+
+    .fa-caret-down {
+      position: absolute;
+      top: 11px;
+      right: 13px;
+      
+      color: $dark-grey;
+    }
```
craigcarlyle: Matching the `thSelect` styling.

themis_components/thFilter/index.scss
JefCurtis: Strange number there but ok.

themis_components/thFilter/index.scss
JefCurtis: Isn't it possible that we might use this styling outside of the filter? Should it be a "condensed" attribute like the rest?

craigcarlyle: I don't think these should belong to the condensed styles, but I could be convinced to move them to a third style for `thAutocomplete`. The only hesitation I have in moving these styles over is that with how it's being used in `thFilter`, it's more of a select/autocomplete hybrid. It seems like more of a relic of the way our legacy components were built rather than the direction we want to go moving forward.

Thoughts?

themis_components/thFilter/index.scss
craigcarlyle: It's the height computed with a list of 10 items! 😄

themis_components/thFilter/index.scss
JefCurtis: That makes sense to me. With Apollo this will have to be rethought anyways.

PR #219, opened 2016-09-21
CLIO-40657: Should return null if no initial value

dtomalty:


themis_components/thFilter/filters/select/selectFilter.service.coffee
```diff
@@ -31,7 +31,7 @@ angular.module "ThemisComponents"
 
     setOption: (value) =>
       @model = @options.find (item) ->
-        item.value?.toString() is value
```
dtomalty: Turns out there's no reason to stringify the value, as we don't stringify it in `getState`. So now we're comparing apples to apples..

themis_components/thFilter/filters/select/selectFilter.service.coffee
```diff
@@ -10,7 +10,7 @@ angular.module "ThemisComponents"
       @valueField = options.selectOptionsValueField or "value"
 
       if options.selectOptionsUrl?
-        @model = value: initialState?.value
+        @model = if initialState then {value: initialState.value} else null
```
seanhealy:
```coffeescript
@model = if initialState? then {value: initialState.value} else null
```

PR #224, opened 2016-09-22
CLIO-40692: Filter loading state initial pass.

dtomalty:


themis_components/thFilter/thFilter.directive.coffee
```diff
@@ -27,10 +27,12 @@ angular.module "ThemisComponents"
     return
 
   link: (scope, element, attrs, thFilterController) ->
+    scope.thFilter.isLoading = true
```
JeffMarvin: Are you sure you don't want to use `yes` and `no`? (troll)

dtomalty: Good comment...didn't even cross my mind!

themis_components/thFilter/examples/7 - Loading State/coffee.coffee
```diff
@@ -0,0 +1,18 @@
+angular.module 'thDemo', ['ThemisComponents']
```
dtomalty: Yah, it occurred to me. This isn't really a suitable thing for an example as it's not something the user will ever have to recreate, it's behaviour that comes free with using the inner directives, it's more for styling. I did struggle for a bit to come up with a better example, but it would have involved changing the way it's implemented for the purposes of writing example code. The only real example would involve writing a directive that would be inside the `<th-filter>` tags. The inner directive would register a promise that would either never resolve or resolve after some duration. But again, it's not something the user would ever have to do. The alternative would be to rip this example out, which is also totally feasible. I just wanted something to show off the loading state, more for debugging etc..

themis_components/thFilter/thFilter.template.html
```diff
@@ -10,4 +10,8 @@
     </th-button>
   </div>
   <div ng-transclude></div>
+
+  <div class="th-filter-overlay" ng-if="thFilter.isLoading">
+    <th-loader class="th-filter-loader" size="small">&nbsp;</th-loader>
```
dtomalty: When left empty or with whitespace, it defaults to the "Loading..." text. This prevents it from displaying any text.

themis_components/thFilter/examples/7 - Loading State/coffee.coffee
JeffMarvin: I think keep it in, but maybe add a note / title to the effect of: this demo is useful for styling purposes, but the behavior it shows is built into thFilter by default.

themis_components/thFilter/thFilter.template.html
JeffMarvin: This is a bit of a hack. Would you consider filing a followup ticket for thLoader to allow this behavior to be triggered by an attribute value?

themis_components/thFilter/thFilter.directive.coffee
```diff
@@ -27,10 +27,12 @@ angular.module "ThemisComponents"
...
     $q.when(
       Promise.all scope.thFilter.initPromises
       ->
+        scope.thFilter.isLoading = false
         scope.thFilter.filterSet.onInitialized?()
       ->
-        console.log 'error'
+        console.log 'thFilter: Some filters were unable to load.'
```
dtomalty: I was going to add an error state, but after talking to @seanhealy I think we're going to hold off until we have a more overarching strategy for handling api errors.

themis_components/thFilter/examples/7 - Loading State/coffee.coffee
dtomalty: Spoke with @seanhealy about this..going to remove the demo as it's not something that's relevant to anyone using the library, it's only purpose is for debugging. We spoke about adding a debug mode to the example browser where it would show examples that are marked as 'debug only'. Will create a ticket for that and move this example into that ticket...

themis_components/thFilter/thFilter.template.html
dtomalty: The relevant line in `thLoader` is `messageElement.innerText ||= "Loading..."`. The `&nbsp` is just used to indicate there's something there, seems like a not perfect but reasonable use case to me.. This was copied over from `thTable`.

themis_components/thFilter/examples/7 - Loading State/coffee.coffee
JeffMarvin: Perfect, good strategy.

PR #229, opened 2016-09-28
Clio 40373 checkbox indeterminate state

JefCurtis:


themis_components/thCheckbox/thCheckbox.directive.spec.coffee
```diff
@@ -61,6 +60,26 @@ describe 'ThemisComponents: Directive: thCheckbox', ->
...
+      scope.$apply -> scope.indeterminate = no
+      expect(element.find("i").hasClass("indeterminate")).toBe false
+
+  describe "when indeterminate state is active and thCheckbox is clicked", ->
+    it "the checked state overrides it", ->
+      compileCheckboxDirective indeterminateTemplate, false, (-> return), true
+      element.triggerHandler "click"
+      expect(element.find("i").hasClass("checked")).toBe true
+      expect(element.find("i").hasClass("indeterminate")).toBe false
+      expect(scope.indeterminate).toBe false
```
seanhealy: We should probably have a test where the checkbox was selected then set to indeterminate then clicked.

themis_components/thCheckbox/thCheckbox.directive.coffee
```diff
@@ -9,17 +9,18 @@ angular.module('ThemisComponents')
...
     controller: ($scope, $element) ->
-      @checked = @checked ? false
 
       @toggle = ->
         $scope.$apply =>
           @checked = not @checked
         @change() if @change?
 
       $element.on 'click', =>
+        @indeterminate = no
```
seanhealy: How do we want the checkbox to behave when leaving the indeterminate state. I feel like it makes the most sense to always leave this state into a selected state. Thoughts?

seanhealy: Perhaps lets talk about this one if you want?

JefCurtis: Shouldn't the checkbox's value just keep the original value? In other words the 'indeterminate' state does _not_ effect the checkbox's state?

PR #243, opened 2016-10-17
CLIO-39529: First pass on thButton accessibility

craigcarlyle:
- thButton: `a` element now has the `role` attribute set to `button`
- thButton: Now has an `aria-disabled` attribute for when it's disabled
- thButton: Can now be triggered with the Enter and Space keys
- thButton: Added accessibility documentation, examples
- thButton: Removed `outline` rule
- thButton: Updated specs to cover the optional aria tags


themis_components/thButton/index.scss
```diff
@@ -34,7 +34,6 @@
 
   border: none;
   border-radius: $border-radius;
-  outline: none;
```
craigcarlyle: This prevents keyboard focus. Big a11y no-no.

themis_components/thButton/thButton.directive.coffee
```diff
@@ -26,5 +28,12 @@ angular.module('ThemisComponents')
       isSubmit = $attrs.submit? or $attrs.type is 'submit'
       type = if isSubmit then 'submit' else 'button'
       $element.attr('type', type)
-      $element.attr('disabled', 'disabled') if $attrs.disabled?
+      if $attrs.disabled?
+        $element.attr('disabled', 'disabled').attr('aria-disabled', true)
```
craigcarlyle: I would prefer this get set in the template, but this allows for `ng-disabled` to be set without any conflicts.

themis_components/thButton/thButton.anchor.template.html
```diff
@@ -1,6 +1,20 @@
...
+    >
+    <th-loader
+      theme="button.theme"
+      size="mini"
+      >
+    </th-loader>
   </span>
-  <ng-transclude ng-class="button.loading ? 'hide' : 'show'"></ng-transclude>
+  <ng-transclude
+    ng-class="button.loading ? 'hide' : 'show'"
```
seanhealy: In this file I'm seeing most the changes as being reformatting of the existing HTML with the addition of role anything else of note?

themis_components/thButton/thButton.button.template.html
```diff
@@ -1,6 +1,19 @@
-<button class="th-button {{button.type}}">
-  <span class="load-wrapper" ng-class="button.loading ? 'show' : 'hide'">
-    <th-loader theme="button.theme" size="mini"></th-loader>
+<button
+  class="th-button {{button.type}}"
+  role="button"
```
seanhealy: Do we need to define the role of a button to be button?

themis_components/thButton/thButton.directive.coffee
seanhealy: Don't quite follow your comment here… What do you mean by conflicts?

themis_components/thButton/thButton.button.template.html
craigcarlyle: Nope, good catch. Will remove this.

themis_components/thButton/thButton.directive.coffee
craigcarlyle: Adding a `ng-disabled` attribute in the template would conflict with a `ng-disabled` attribute included in usage:

Template:
`<button ng-disable="foo"><ng-transclude></<ng-transclude></button>`

Usage:
`<th-button ng-disable="bar">Submit</th-button>`

Can those two ng-disable properties be combined in the controller? Or is this approach fine?

frankieyan: Never used before but would this help? https://docs.angularjs.org/api/ngAria

PANCAKES: That's really neat, @frankieyan!

seanhealy: We can have both together. Much the same way we handle the ng-change and other similar properties.

seanhealy: I'm a little bit confused as to where we ended up on this re the original comment. Wanna ping me tomorrow at some point and we can chat over it really quickly?

PR #254, opened 2016-10-24
CLIO-39551: thTabset a11y

craigcarlyle:


themis_components/thTabset/thTabset.directive.coffee
```diff
@@ -40,4 +42,41 @@ angular
         $scope.activateTab tab if tabs.length is 0
         tabs.push tab
 
+      $element.on 'keyup', (event) ->
+        $timeout ->
+          if angular.element(event.target).hasClass('tab-name')
+            if event.keyCode == keycode('Right')
```
PANCAKES: How does this interact in a larger scope? I mean, if this is in a page and I am tabbing through the controls/elements on the page, I will end up on the first tab in normal tab order? I guess the first tab will be showing its content, so subsequent tab keypresses will move us from the tabset to the next control and left will move us to the next tab? Mostly just trying to work out the bigger workflow in my mind - not saying you have done anything wrong.

themis_components/thTabset/thTabset.directive.coffee
```diff
@@ -40,4 +42,41 @@ angular
...
+                else
+                  i++
+
+            if event.keyCode == keycode('Left')
+              event.preventDefault()
+              i = 0
+              while i < $scope.tabs.length
+                if $scope.tabs[i].active
+                  if i == 0
+                    lastTab = $element.find('li')[$scope.tabs.length - 1]
```
PANCAKES: This will cycle us to the last tab in the set, if we press left while on the first tab?

themis_components/thTabset/thTabset.directive.coffee
```diff
@@ -40,4 +42,41 @@ angular
...
+      $element.on 'keyup', (event) ->
+        $timeout ->
+          if angular.element(event.target).hasClass('tab-name')
+            if event.keyCode == keycode('Right')
+              event.preventDefault()
+              i = 0
+              while i < $scope.tabs.length
+                if $scope.tabs[i].active
+                  if i == $scope.tabs.length - 1
+                    firstTab = $element.find('li')[0]
```
PANCAKES: This will cycle us to the first tab if we press right arrow while on the last tab?

themis_components/thTabset/thTabset.directive.spec.coffee
```diff
@@ -48,16 +48,16 @@ describe 'ThemisComponents: Directive: thTabset', ->
       expect(element.find('.th-tab-bar').length).toBe 1
 
     it 'should create a content area', ->
-      expect(element.find('.th-tabset-content').length).toBe 1
+      expect(element.find('#th-tabset-content').length).toBe 1
```
PANCAKES: Is this ID only used for testing, or is it used for the actual deployed component? I worry if we have 2 tabsets on a page that we will have multiple components using the same ID.

themis_components/thTabset/thTabset.native.template.html
```diff
@@ -1,25 +1,37 @@
...
+    <div
+      class="th-tab-action-bar"
+      ng-transclude="actionBar"
+      tabindex="-1"
+    >
+    </div>
+  </ul>
   <div
-    class="th-tabset-content"
+    id="th-tabset-content"
```
PANCAKES: This guy here - if we have 2 tabsets on the page, we will be in trouble, no?

themis_components/thTabset/thTabset.directive.coffee
craigcarlyle: It ends up on the active tab by default. 

Pressing tab goes to the next element.

The left and right arrow keys go to the next tab (as described in the WAI guide).

themis_components/thTabset/thTabset.directive.coffee
craigcarlyle: Yes.

themis_components/thTabset/thTabset.directive.coffee
craigcarlyle: Yes.

themis_components/thTabset/thTabset.native.template.html
craigcarlyle: This change is no longer needed. Resolved.

themis_components/thTabset/thTabset.directive.spec.coffee
craigcarlyle: This change is no longer needed. Resolved.

PR #257, opened 2016-10-26
CLIO-41549: Fix for bug where value is set to null when user selects …

dtomalty:
…one of the initial options.


themis_components/thAutocomplete/thAutocomplete.directive.coffee
```diff
@@ -98,8 +98,20 @@ angular.module "ThemisComponents"
...
+
+      # This clears the autocomplete (in single-selection only) when the
+      # user clears the input text
+      checkForEmptySearchField = ->
+        if not multiple
+          search = angular.element(element[0].querySelectorAll(".ui-select-search"))
+          lastSearchIsEmpty = not (controller.lastSearch?.length > 0)
+          thisSearchIsEmpty = not (search.val()?.length > 0)
+
+          if not lastSearchIsEmpty and thisSearchIsEmpty
```
JefCurtis: Can we put a comment saying something like:
```coffescript
# We only want to set the model as null when there is actually something to clear.
# If it was already empty then don't clear it again.
```

PR #259, opened 2016-10-27
Clio 41539 hide empty custom field filter + migrate to component

frankieyan:


themis_components/thFilter/thCustomFilters.component.coffee
```diff
@@ -0,0 +1,88 @@
+class CustomFilters
```
frankieyan: Sucks that git won't pick up the rename here - the same thing would happen when we switch to TS too.

themis_components/thFilter/thCustomFilters.component.coffee
```diff
@@ -0,0 +1,88 @@
...
+    identifier = @_nextIdentifier
+    @_nextIdentifier += 1
+    identifier.toString()
+
+angular.module "ThemisComponents"
+.component "thCustomFilters",
+  require:
+    thFilterCtrl: "^thFilter"
+  bindings:
+    options: "<"
```
dtomalty: What does this do?

themis_components/thFilter/thCustomFilters.component.coffee
```diff
@@ -0,0 +1,88 @@
...
+          item.fieldIdentifier is key
+
+        unless type? then continue
+
+        @customFilterRows.push
+          identifier: @_getNextIdentifier()
+          type: type
+          initialState: @initialState[key]
+
+  _getNextIdentifier: ->
```
dtomalty: Why the underscore?

themis_components/thFilter/thCustomFilters.component.coffee
```diff
@@ -0,0 +1,88 @@
...
+
+  $onInit: ->
+    {
+      filterSet
+      customFilterTypes
+      customFilterUrl
+      customFilterConverter
+      initialState
+      name
+    } = @options or @thFilterCtrl.options
```
dtomalty: Ahh I see...from the binding below.

themis_components/thFilter/thCustomFilters.component.spec.coffee
```diff
@@ -0,0 +1,226 @@
...
+  compileDirective
+} = require "spec_helpers"
+
+describe "ThemisComponents: Component: thCustomFilters", ->
+  FilterSet = filterSet = CustomFilterConverter = httpBackend = thFilterCtrl = $componentController = null
+
+  createFilterCtrl = -> registerInitPromise: jasmine.createSpy "registerInitPromise"
+  createController = (bindings) ->
+    controller = $componentController "thCustomFilters", null, bindings
+    controller.thFilterCtrl = createFilterCtrl()
```
dtomalty: Not sure what's happening here..

themis_components/thFilter/thCustomFilters.component.spec.coffee
```diff
@@ -0,0 +1,226 @@
+{
+  compileDirective
+} = require "spec_helpers"
+
+describe "ThemisComponents: Component: thCustomFilters", ->
+  FilterSet = filterSet = CustomFilterConverter = httpBackend = thFilterCtrl = $componentController = null
```
dtomalty: Just a general comment..I love that we don't have to instantiate using a template :)

themis_components/thFilter/thCustomFilters.component.spec.coffee
```diff
@@ -0,0 +1,226 @@
...
+        filterSet: filterSet
+        customFilterTypes: []
+
+      controller = $componentController "thCustomFilters", null, null
+      controller.thFilterCtrl = thFilterCtrl
+      controller.$onInit()
+
+      expect(controller.filterSet).toBe filterSet
+
+  describe "when filter set is defined on th-custom-filters", ->
```
dtomalty: I wonder if we should maintain being able to set `filterSet` on `th-custom-filters`?

themis_components/thFilter/thCustomFilters.component.spec.coffee
```diff
@@ -0,0 +1,226 @@
...
+    endpoint = -> "/custom_filter.json"
+
+    beforeEach ->
+      results = [
+        {type: "input", fieldIdentifier: "id0", name: "input"}
+        {type: "select", fieldIdentifier: "id1", name: "select"}
+      ]
+
+      httpBackend.when "GET", endpoint
+        .respond JSON.stringify results
```
dtomalty: Much cleaner..

themis_components/thFilter/thCustomFilters.component.spec.coffee
```diff
@@ -0,0 +1,226 @@
...
+      controller.$onInit()
+      controller.addCustomFilterRow()
+      controller.addCustomFilterRow()
+
+    describe "when called with invalid index", ->
+      it "should throw an error", ->
+        expect(-> controller.removeCustomFilterRow lastIndex + 1).toThrow()
+
+    describe "when called with index of first element repeatedly", ->
+      it "should remove first element from DOM repeatedly", ->
```
dtomalty: Maybe remove references to the DOM from here and the test below.

themis_components/thFilter/thCustomFilters.component.coffee
```diff
@@ -0,0 +1,88 @@
+class CustomFilters
+  constructor: (@CustomFilterConverter, @$http, @$timeout) ->
+    @customFilterRows = []
+    @_nextIdentifier = 0
```
frankieyan: Not really an Angular 2 thing. This wasn't an exposed property before, so I figured we should mark it as internal/private. When we convert to TS we can write it as `private nextIdentifer = 0`

themis_components/thFilter/thCustomFilters.component.coffee
frankieyan: For noting that its a private method

themis_components/thFilter/thCustomFilters.component.coffee
frankieyan: One way binding

themis_components/thFilter/thCustomFilters.component.spec.coffee
frankieyan: But we lose some of the template specific tests!

themis_components/thFilter/thCustomFilters.component.spec.coffee
frankieyan: It's returning a mocked Filter controller (the required parent controller) like `{ registerInitPromise: spy() }`. I think we're not actually spying on that method though but we *could* just to check if we're properly registering our promise with the parent.

themis_components/thFilter/thCustomFilters.component.spec.coffee
frankieyan: Do we ever want to override the parent's filterset? Since we've marked the parent filter component as required, this component can't be standalone anymore.

themis_components/thFilter/thCustomFilters.component.spec.coffee
frankieyan: Added additional test to cover that ^

themis_components/thFilter/readme.md
```diff
@@ -299,16 +277,24 @@ hash consisting of the following options:
             )()
     ```
 
+The follow property is supported when the `th-search-row` component is included as a child element of `th-filter`:
```
dtomalty: "following"

themis_components/thFilter/thCustomFilters.component.spec.coffee
```diff
@@ -86,24 +78,38 @@ describe "ThemisComponents: Component: thCustomFilters", ->
...
 
     it "should set customFilterTypes to response", ->
-      expect(controller.customFilterTypes.length).toBe results.length
-      expect(controller.customFilterTypes[0].type).toBe results[0].type
-      expect(controller.customFilterTypes[0].fieldIdentifier).toBe results[0].fieldIdentifier
-      expect(controller.customFilterTypes[0].name).toBe results[0].name
-      expect(controller.customFilterTypes[1].type).toBe results[1].type
-      expect(controller.customFilterTypes[1].fieldIdentifier).toBe results[1].fieldIdentifier
-      expect(controller.customFilterTypes[1].name).toBe results[1].name
+      response = results()
```
dtomalty: What's the reason for making `results` a function?

themis_components/thFilter/thCustomFilters.component.spec.coffee
```diff
@@ -86,24 +78,38 @@ describe "ThemisComponents: Component: thCustomFilters", ->
...
 
     it "should add a row for each initial known name/value pair", ->
+      httpBackend.whenGET(endpoint()).respond JSON.stringify results()
+      httpBackend.flush()
+
       expect(controller.customFilterRows.length).toBe 1
       expect(controller.customFilterRows[0].initialState).toEqual {value: "test"}
 
+    it "should resolve the promise registered with thFilter", (done) ->
+      promise = controller.thFilterCtrl.registerInitPromise.calls.argsFor(0)[0]
```
dtomalty: Is there nothing to `expect` in this test? We're just checking that it doesn't throw any errors now?

themis_components/thFilter/thCustomFilters.component.spec.coffee
```diff
@@ -86,24 +78,38 @@ describe "ThemisComponents: Component: thCustomFilters", ->
...
+      promise.then -> done()
```
dtomalty: What is `done()`?

themis_components/thFilter/thCustomFilters.component.spec.coffee
```diff
@@ -134,25 +140,31 @@ describe "ThemisComponents: Component: thCustomFilters", ->
               fieldIdentifier: idPrefix() + item.id
               name: namePrefix() + item.id
               type: typePrefix() + item.id
-        testConverter = new TestCustomFilterConverter
+        testConverter = new TestCustomFilterConverter()
```
dtomalty: Is this strictly necessary?

themis_components/thFilter/thCustomFilters.component.spec.coffee
dtomalty: Ahh nm I see now. Does the test not finish if `done` isn't called?

themis_components/thFilter/thCustomFilters.component.spec.coffee
frankieyan: It's just to make the response array immutable and we have a fresh copy everytime.. we could either do this or set it in beforeEach.

themis_components/thFilter/thCustomFilters.component.spec.coffee
frankieyan: Just checking that the promise is resolved. If the `.then` and `done()` doesn't get fired this test will fail.

themis_components/thFilter/thCustomFilters.component.spec.coffee
frankieyan: Not necessary but like it for consistency. Can't find a specific style-guide rule on this but: https://github.com/airbnb/javascript#variables--const and https://github.com/airbnb/javascript#constructors--chaining shows usage with parenthesis in place.

dtomalty: I think I prefer it anyway as it emphasizes that a new object is being created.

themis_components/thFilter/thCustomFilters.component.spec.coffee
dtomalty: Right. Looks good.

PR #262, opened 2016-11-01
Clio 40755 datepicker filter tests

JeffMarvin:


package.json
```diff
@@ -77,7 +77,7 @@
     "clean": "rm -rf public/build",
     "postinstall": "npm run clean && npm run build",
     "start": "gulp docs",
-    "test": "karma start",
+    "test": "TZ='UTC' karma start",
```
JeffMarvin: This sets the environment variable for the tests run in PhantomJS to use UTC as the timezone. This is good to lock down, as CI and other environments are in different timezones.

themis_components/thFilter/filters/date/dateFilter.service.coffee
```diff
@@ -14,11 +14,21 @@ angular.module "ThemisComponents"
 
       @clearState()
 
+      if @operatorOptions?
```
JeffMarvin: Adds some additional checking to the dateFilter service. It can be used with and without OperatorOptions, so this provides stricter checking if it does.

themis_components/thFilter/filters/date/thFilter.date.directive.coffee
```diff
@@ -6,9 +6,9 @@ angular.module "ThemisComponents"
     scope:
       filterSet: "="
       filterOptions: "="
-      operatorOptions: "="
-      defaultOperatorIndex: "@"
-      placeholder: "@"
+      operatorOptions: "=?"
```
JeffMarvin: More accurately, these are optional in the directive.

themis_components/thFilter/filters/date/dateFilter.service.spec.coffee
```diff
@@ -0,0 +1,118 @@
...
+    it "should inherit from FilterBase", ->
+      expect(dateFilter.prototype).toBe instanceof FilterBase
+
+    describe "when operatorOptions specified", ->
+      describe "and initial state specified", ->
+        beforeEach ->
+          dateFilter = new DateFilter(
+            {placeholder: "placeholder"}
+            [{name: "a", value: "b"}, {name: "c", value: "d"}]
+            2
```
PANCAKES: I expected this to throw in initialization. If our index is 2, but we have 2 items in the array, we index outside our items if we access item [2].

themis_components/thFilter/filters/date/dateFilter.service.coffee
```diff
@@ -14,11 +14,21 @@ angular.module "ThemisComponents"
...
+        unless @operatorOptions instanceof Array and @operatorOptions.length > 0
+          throw new Error "DateFilter: must specify operator options of type Array."
+
+        if @defaultOperatorIndex? and @defaultOperatorIndex > @operatorOptions.length
```
PANCAKES: Should we throw if the ```@defaultOperatorIndex``` is >= ```@operatorOptions.length```?

JeffMarvin: Yes, yes we should. Fixed this and the subsequent test below that you commented on.

PR #270, opened 2016-11-03
Clio 41684 subtext th input

acorcuera:


themis_components/thWithLabel/readme.md
```diff
@@ -32,7 +32,9 @@ in conjunction with the `with-subtext` component.
 - It can only be used as an attribute in conjunction with the `with-label` component.
 - Existing labels for input types `checkbox` and `radio`,
 will be moved to the top of their container, and the new label for the subtext will appear below.
-- It can only be used with `thCheckbox` and `thRadioButton` components,
+- The new label for the subtext for input type `text` will appear below the input component.
```
acorcuera: @JefCurtis , it's showing up at the bottom of the subtext

themis_components/thWithLabel/index.scss
```diff
@@ -21,6 +21,25 @@ label.th-label {
     }
   }
 
+  .sublabel-text {
```
JefCurtis: when he ignores that black kid _e2719f0ad77fbe5595ce3ec12d5cb92c

themis_components/thWithLabel/readme.md
```diff
@@ -32,7 +32,9 @@ in conjunction with the `with-subtext` component.
 - It can only be used as an attribute in conjunction with the `with-label` component.
 - Existing labels for input types `checkbox` and `radio`,
 will be moved to the top of their container, and the new label for the subtext will appear below.
-- It can only be used with `thCheckbox` and `thRadioButton` components,
```
JefCurtis: This is not quite clear. Maybe something like. I think we need to changes this section entirely. Something like:
> - It can only be used as an attribute in conjunction with the `with-label` component.
- It currently works with `thCheckbox`, `thRadio`, and `thInput`.
- When added to a inline elements such as `thCheckbox` and `thRadio` it will:
  - move the existing labe to the top of their container, and the subtext will appear below it.
- When added to a block elements such as `thInput` it will:
  - position the subtext directly below the input component.

themis_components/thWithLabel/sharedTests.coffee
```diff
@@ -24,6 +24,20 @@ module.exports =
     it "adds 'with-subtext' value to nested sub label", ->
       expect(element.next().text()).toMatch "label subtext"
 
+  testingBlockSubLabel: (compileElementFn) ->
+    element = null
+    beforeEach ->
+      {element} = compileElementFn()
+
+    it "adds subtext class to container", ->
+      expect(element.is("span.th-input-wrapper.ng-scope.ng-isolate-scope.with-subtext")).toBe true
```
JefCurtis: we should only be testing that `element.is("span.with-subtext")`. The rest of those classes are irrelevant. Also I know this is not your change but can you update line 19 as well. 
tumblr_m4ykmv3j4i1qb4t0ro1_250

themis_components/thWithLabel/thWithSubtext.directive.coffee
```diff
@@ -37,3 +33,11 @@ angular.module("ThemisComponents")
           span = label[0].getElementsByClassName "label-text"
           label.children().addClass 'with-subtext'
           span[0].innerHTML += "<p class='inline sublabel-text'>#{attrs.withSubtext}</p>"
+        else
+          label.addClass 'subtext-label'
+          subtextParagraph =
+            angular.element("<p class='block sublabel-text'>#{attrs.withSubtext}</p>")
```
JefCurtis: here as well.

PR #273, opened 2016-11-06
CLIO-41775 Exclude polyfills and specs from coverage report

frankieyan:


karma.conf.coffee
```diff
@@ -95,7 +95,15 @@ module.exports = (config) ->
       transform: [
         'coffeeify'
         'stringify'
-        coverage
+        coverage {
+          ignore: [
+            "**/*.html"
+            "**/*.spec.*"
+            "**/polyfills/**"
+            "**/public/javascript/**"
```
seanhealy: We can probably just ignore everything in `public`.

PR #274, opened 2016-11-08
Clio 41276 disclosure animation fix

frankieyan:


themis_components/services/utilities.service.coffee
```diff
@@ -3,6 +3,17 @@ angular.module 'ThemisComponents'
...
+      previousCss = element.getAttribute "style"
+      element.style.position = "absolute"
+      element.style.visibility = "hidden"
+      element.style.maxHeight = "none"
+      height = element.offsetHeight
+
+      element.setAttribute "style", previousCss or ""
+      return height
+
+    return {
```
frankieyan: Removed Object.freeze so that these methods can be spied on - do we have problems with these being overwritten?

themis_components/thDisclosure/thDisclosureContent.directive.coffee
```diff
@@ -1,58 +1,34 @@
...
-
-close = (element) ->
-  $(element).css overflow: 'hidden'
-
-angular.module 'ThemisComponents'
-  .directive 'thDisclosureContent', (DisclosureManager) ->
-    restrict: 'E'
+class DisclosureContent
+  constructor: ($element, $timeout, DisclosureManager, Utilities) ->
+    @_disclosureManager = DisclosureManager
```
frankieyan: I know these _ prefixes are ugly 😛 but we want to remember to mark them as private when we migrate to TypeScript

themis_components/services/utilities.service.coffee
mbuckley: Should we guard against element being null here?

PR #278, opened 2016-11-10
Clio 41821 add defaults

seanhealy:


themis_components/thDefaults/thDefaults.service.coffee
```diff
@@ -0,0 +1,37 @@
+angular.module 'ThemisComponents'
```
mbuckley: Did we look at making this a provider? It seems to fit the purpose and might be a nice way to config libThemisUI once we have more than a few config options.

```
themisApp.config(function($libThemisUIProvider) {
  $libThemisUIProvider.set "dateFormat", "DD/MM/YYYY"
}]);
```

More a thought..might be overkill. What do you think?

PR #279, opened 2016-11-14
Clio 41276 disclosure fix

frankieyan:


themis_components/thDisclosure/index.scss
```diff
@@ -1,5 +1,5 @@
 th-disclosure-content {
-  ng-transclude {
+  > ng-transclude {
```
seanhealy: Should this be `& > ng-transclude`?

PR #284, opened 2016-11-16
clio-41592: Hide "Select All" when there is nothing to select.

acorcuera:


themis_components/thActionBar/index.scss
```diff
@@ -23,6 +23,11 @@
         cursor: pointer;
       }
     }
+
+    &.hidden {
```
seanhealy: I think I know the answer from an old convo with Jef but, it might be worth documenting in a comment or something why you can't just use `ng-if` or `ng-show` here.

PR #298, opened 2016-11-23
Clio 41276 disclosure fix retry

frankieyan:
Changes vs last implementation is at https://github.com/frankieyan/lib-themisui/blob/ea826fbac59e93633aa5f0a128cf579e3ff954a0/themis_components/thDisclosure/thDisclosureContent.component.coffee#L36 where it was incorrectly setting the max-height to the content's height before setting it to 0 again, when initializing the component with a closed state.


themis_components/thDisclosure/index.scss
```diff
@@ -1,8 +1,14 @@
 th-disclosure-content {
-  width: 100%;
-  overflow: hidden;
-  height: 0;
-  display: block;
+  > ng-transclude {
+    width: 100%;
+    display: block;
+    overflow: hidden;
+    transition: max-height .3s;
```
craigcarlyle: This is super nit-picky, but I believe there should be a blank line separating the `overflow` and `transition` declarations (according to our [CSS Declaration Order](https://themis.atlassian.net/wiki/display/PDEV/CSS+Declaration+Order)).

themis_components/thDisclosure/thDisclosureContent.component.coffee
```diff
@@ -0,0 +1,50 @@
...
+          @open = yes
+        , @_transitionTimeout
+    else
+      @_$timeout =>
+        @open = no
+
+        if @contentHeight > 0
+          @contentHeight = height
+
+        @_$timeout =>
```
craigcarlyle: This seems a little hacky, no?

themis_components/thDisclosure/thDisclosureToggle.template.html
```diff
@@ -1,6 +1,6 @@
 <a href ng-click="thDisclosureToggle.toggle()">
   <ng-transclude
-    ng-if="thDisclosureToggle.textSide == 'left'"
+    ng-if="thDisclosureToggle.textSide === 'left'"
```
craigcarlyle: Why was this change required?

themis_components/thDisclosure/thDisclosureContent.component.coffee
frankieyan: Super hacky - every browser other than Safari was able to handle the max-height transition fine. Will investigate some more

frankieyan: ^ Also discovered a race condition when spamming the toggle where open would always beat out close

themis_components/thDisclosure/thDisclosureToggle.template.html
frankieyan: Not required just a good habit to get into especially when we will migrate from coffeescript (Coffeescript compiles == into ===)

PR #299, opened 2016-11-24
CLIO-41709: Fix error thrown when null values are included in thTable

craigcarlyle:


themis_components/thTable/thTableSort.service.coffee
```diff
@@ -9,10 +9,12 @@ TableSort = ->
       -compareResult
 
   compare = (a, b) ->
-    if typeof a in ["number", "boolean"]
+    if typeof a == "number"
```
JeffMarvin: Technically should be `is` not `==` by our style guide. Same with below (15-17).

craigcarlyle: Addressed.

themis_components/thTable/thTableSort.service.coffee
```diff
@@ -9,10 +9,12 @@ TableSort = ->
       -compareResult
 
   compare = (a, b) ->
-    if typeof a in ["number", "boolean"]
+    if typeof a is "number"
       a - b
     else
-      a.localeCompare b
+      va = if a is null then '' else '' + a
```
seanhealy: Also can we use variable names here that are a bit more descriptive. This is a bit hard to follow.

themis_components/thTable/thTableSort.service.coffee
```diff
@@ -9,10 +9,12 @@ TableSort = ->
...
+      vb = if b is null then '' else '' + b
+      if va > vb then 1 else if va is vb then 0 else -1
```
seanhealy: We should keep using `localeCompare` as it deals with a lot of edge cases for us.

PR #302, opened 2016-11-29
CLIO-42363: Add condensed styling option to subtabs.

dtomalty:


themis_components/thTabset/thTab.directive.coffee
```diff
@@ -9,6 +9,7 @@ angular
       name: "@name"
       ngClick: "&"
       badge: "<"
+      letterSpacing: "=?"
```
frankieyan: Can we make this one-way bound `"<?"`? I don't think we need to explicitly mark this as optional here either as it'd just be `undefined` or falsy when not passed in.

PR #305, opened 2016-12-05
Add coffeelint rules

frankieyan:
* Install Coffeelint so gulp-coffeelint will pick up the newer version as opposed to the one in it's dependency - This gets rid of the @ stand alone false positive.
* Added the following rules which are set to `warn`:
  * Cyclomatic complexity
  * English operator (and, or, is, isnt)
  * Double quotes
  * Omit parenthesis where possible
  * Check for fit/fdescribe


coffeelint.json
```diff
@@ -56,11 +60,30 @@
...
     "level": "error"
   },
 
   "spacing_after_comma": {
     "level": "error"
+  },
+
+  "no_focused_jasmine_specs": {
+    "module": "coffeelint-no-focused-jasmine-specs",
+    "level": "warn"
```
PANCAKES: Should we set this to "error"?

seanhealy: Probably a good idea.

PR #318, opened 2016-12-09
Use full github url for datepicker

mbuckley:
Due to an issue with npm shrinkwrap and installing on the themis side, we have updated datepicker to use a full url to github.


seanhealy:
👍

PR #320, opened 2016-12-15
CLIO-42190 thTable hover highlight fix

craigcarlyle:


src/lib/thTable/thTable.directive.coffee
```diff
@@ -7,9 +7,31 @@ angular.module 'ThemisComponents'
       delegate: '='
     bindToController: true
     controllerAs: 'thTable'
-    controller: ($scope, $element) ->
+    controller: ($scope, $element, $timeout) ->
       thTableKeyboardNavigation $element, $scope
 
+      @mouseOver = (event) ->
+        $timeout ->
```
JeffMarvin: I am curious, what is timeout needed for here?

src/lib/thTable/thTable.directive.coffee
```diff
@@ -7,9 +7,31 @@ angular.module 'ThemisComponents'
...
+          eventRows = getEventRows event
+          angular.element(eventRows.hoverRow).addClass 'th-table-hover-row'
```
JeffMarvin: Here and elsewhere in the PR, styleguide sez double quotes by default.

src/lib/thTable/thTable.directive.coffee
```diff
@@ -7,9 +7,31 @@ angular.module 'ThemisComponents'
...
+          angular.element(eventRows.partnerRow).addClass 'th-table-hover-row'
+
+      @mouseLeave = (event) ->
+        $timeout ->
+          eventRows = getEventRows event
+          angular.element(eventRows.hoverRow).removeClass 'th-table-hover-row'
+          angular.element(eventRows.partnerRow).removeClass 'th-table-hover-row'
+
+      getEventRows = (event) ->
```
JeffMarvin: I see how this logic fixes the bug, but this treatment feels brittle. In the future, how much do you think we will want to generalize this to different types of rows? Any thoughts on a way to, maybe, pass a set or bind a list of rows that behave together? I say this without a full solution in mind, just thinking out loud....

src/lib/thTable/thTable.directive.spec.coffee
```diff
@@ -81,3 +82,47 @@ describe "ThemisComponents: Directive: thTable", ->
                     .text()
                     .trim()
     expect(renderedData).toEqual outsider
+
```
JeffMarvin: Woo! tests! 🎆

src/lib/thTable/thTable.directive.coffee
craigcarlyle: Addressed.

src/lib/thTable/thTable.directive.coffee
craigcarlyle: Addressed.

src/lib/thTable/thTable.directive.coffee
JeffMarvin: Yup, makes sense. Especially sold on the fact that we are working on a new component for tables. Let's remember to make it better next time. ;)

src/lib/thTable/thTable.directive.spec.coffee
```diff
@@ -7,9 +7,10 @@ describe "ThemisComponents: Directive: thTable", ->
   SimpleTableDelegate = TableHeader = element = null
 
   beforeEach angular.mock.module 'ThemisComponents'
-  beforeEach inject (_SimpleTableDelegate_, _TableHeader_) ->
+  beforeEach inject (_SimpleTableDelegate_, _TableHeader_, $timeout) ->
```
JeffMarvin: This can go now, as well as line 13 declaring `timeout`.

PR #326, opened 2016-12-21
clio-42850: Add show attribute to thTab

JefCurtis:


src/lib/thTabset/thTabset.directive.spec.coffee
```diff
@@ -71,6 +78,19 @@ describe 'ThemisComponents: Directive: thTabset', ->
       expect(element.find('.th-tab-bar li.tab-name').first().hasClass('active')).toBe false
       expect(element.find('.th-tab-bar li.tab-name').last().hasClass('active')).toBe true
 
+  describe "with a show attribute", ->
+    beforeEach ->
+      additions = show: false
+      {element, scope} = compileDirective showTemplate, additions
+
+    it "should add/remove the tab set conditionally with the show attribute", ->
+      tabs = element.find(".th-tab-bar li.tab-name")
```
dtomalty: No parentheses.

src/lib/thTabset/thTabset.directive.spec.coffee
```diff
@@ -71,6 +78,19 @@ describe 'ThemisComponents: Directive: thTabset', ->
...
+      expect(tabs.length).toEqual 1
+
+      scope.$apply -> scope.show = true
+      tabs = element.find(".th-tab-bar li.tab-name")
```
dtomalty: No parentheses.

src/lib/thTabset/thTab.directive.coffee
```diff
@@ -8,9 +8,12 @@ angular
     scope:
       name: "@name"
       ngClick: "&"
+      show: "=?"
```
dtomalty: @frankieyan should this have a question mark?

src/lib/thTabset/thTabset.directive.spec.coffee
```diff
@@ -51,6 +51,13 @@ describe 'ThemisComponents: Directive: thTabset', ->
     </div>
   """
 
+  showTemplate = """
+    <div th-tabset>
```
dtomalty: What if we set `active-tab` to a tab that isn't showing? Looks like it'll still enable the content, just won't show the tab itself?

dtomalty: What if we flip show on the tab that's active? It's just going to hide the tab itself and not the content, right?

src/lib/thTabset/thTabset.native.template.html
```diff
@@ -15,6 +15,7 @@
       aria-controls="{{tab.ariaControlsID}}"
       role="tab"
       aria-selected="{{tab.active}}"
+      ng-if="tab.show"
```
dtomalty: So this applies only to the actual tab..I forget what this is being used for..

src/lib/thTabset/thTab.directive.coffee
frankieyan: I don't think angular blows up if we don't provide an attribute not marked as optional, so it's probably unnecessary. Is the consumer of the tab directive expected to watch the value of show as it changes (when it gets the default value set below)? If not - can we make this a one-way binding?

src/lib/thTabset/thTabset.directive.spec.coffee
JefCurtis: Ok so I've updated it to include the content as well as the tab itself. If the tab is not showing I didn't think it would be necessary to hide the content as well considering it couldn't be "activated" without a tab to click. If the tab is active it will still not show either the content or the tab.

src/lib/thTabset/thTab.directive.coffee
JefCurtis: Switched to one way binding. 👍

src/lib/thTabset/thTabset.native.template.html
JefCurtis: The `show` value now hides both the content and the tab itself.

src/lib/thTabset/thTabset.directive.coffee
```diff
@@ -38,6 +38,43 @@ angular
...
+          tabToActivate = checkForVisibleTabsRightToLeft idx
+          $scope.processTabChange tabToActivate if tabToActivate
+
+        # If there is NOT a tab to the left we'll then check the right side.
+        if idx < (tabs.length - 1) and tabRemoved.active
+          tabToActivate = checkForVisibleTabsLeftToRight idx
+          $scope.processTabChange tabToActivate if tabToActivate
+
+      checkForVisibleTabsRightToLeft = (idx) ->
+        for index in [(idx - 1) .. 0]
```
dtomalty: Maybe just
```
for index in [(idx - 1) .. 0]
  tabs[index] if tabs[index].show
```

src/lib/thTabset/thTabset.directive.coffee
```diff
@@ -38,6 +38,43 @@ angular
...
+          nextLeftTab = tabs[index]
+
+          return nextLeftTab if nextLeftTab.show
+
+      checkForVisibleTabsLeftToRight = (idx) ->
+        for index in [(idx + 1) .. (tabs.length - 1)]
```
dtomalty: Same as above.

src/lib/thTabset/thTabset.directive.spec.coffee
```diff
@@ -71,6 +78,31 @@ describe 'ThemisComponents: Directive: thTabset', ->
...
+      tabs = element.find ".th-tab-bar li.tab-name"
+
+    it "should add/remove the tab set conditionally with the show attribute", ->
+      expect(tabs.length).toEqual 2
+
+      scope.$apply -> scope.show = false
+      tabs = element.find ".th-tab-bar li.tab-name"
+      expect(tabs.length).toEqual 1
+
+    it "should move active tab to next tab if another tab is visible", ->
```
dtomalty: Nice test!

PR #332, opened 2016-12-28
CLIO-42793: Hide thSelect placeholder if not used

craigcarlyle:


tools/gulp/tasks/docsServer.coffee
```diff
@@ -96,6 +96,7 @@ gulp.task 'docs-server', ->
   app.use "/public", express.static 'src/docs-app'
   app.use "/build", express.static 'dist'
   app.use "/exampleTemplates", express.static 'src/docs-app/exampleTemplates'
+  app.use "/json", express.static 'src/docs-app/json'
```
craigcarlyle: This fixes Example 6 of thFilter.

src/lib/thSelect/thSelect.directive.coffee
```diff
@@ -85,6 +85,15 @@ angular.module('ThemisComponents')
       scope.$watch ->
         controller.ngModel
       , (newValue) ->
+
+        isPlaceholder = newValue?
+        # Clear any other selected options
```
craigcarlyle: Discovered this bug while working on this ticket: 

Currently thSelect adds a `selected` attribute to any option element that gets selected. However it never gets cleared, so if you select all 6 options, all 6 options will end up having the `selected` attribute.

This clears other selections when the model is updated.

src/lib/thSelect/thSelect.template.html
```diff
@@ -22,7 +22,7 @@
     ng-required="select.ngRequired"
     >
     <option
-      ng-show="select.placeholder"
```
craigcarlyle: ngShow elements remain in the DOM, which is why thFilter select elements were grabbing the wrong option.

src/lib/thSelect/thSelect.directive.coffee
```diff
@@ -57,7 +57,15 @@ angular.module('ThemisComponents')
...
             @ngModel = {name: element.text, value: element.value}
 
-      @thOnChange = -> Utilities.onChange @ngChange
+      @thOnChange = ->
+        Utilities.onChange @ngChange
+
+        # Clear any other selected options
+        optionElements = $element.find("option")
+        for optionElement in optionElements
+          unless parseInt(@ngModel?.value) == optionElement.id
```
JeffMarvin: `==` should be `is` according to our style guide.

PR #337, opened 2017-01-04
calling 'getState' on a selectFilter will now return the name as well.

JefCurtis:


src/lib/thFilter/filters/select/selectFilter.service.coffee
```diff
@@ -10,10 +10,10 @@ angular.module "ThemisComponents"
       @valueField = options.selectOptionsValueField or "value"
 
       if options.selectOptionsUrl?
-        @model = if initialState? then {value: initialState.value} else null
+        @model = if initialState? then name: initialState.name, value: initialState.value else null
```
dtomalty:
```
@model =
  if initialState?
    name: initialState.name
    value: initialState.value 
  else null
```

src/lib/thFilter/filters/select/thFilter.select.directive.spec.coffee
```diff
@@ -67,7 +67,7 @@ describe "ThemisComponents: Directive: thFilterSelect", ->
       select.val options[1].value
       select.triggerHandler "change"
       timeout.flush()
-      expect(filterSet[0].getState()).toEqual {value: options[1].value}
+      expect(filterSet[0].getState()).toEqual name: options[1].name, value: options[1].value
```
dtomalty:
```
expect(filterSet[0].getState()).toEqual 
  name: options[1].name
  value: options[1].value
```

src/lib/thFilter/filters/select/selectFilter.service.coffee
```diff
@@ -35,7 +35,11 @@ angular.module "ThemisComponents"
 
     getState: =>
       if @model
-        return value: @model.value
+        return {
```
JefCurtis: This doesn't seem to work the same when returning directly. Only when assigning an object.

JefCurtis: Can you confirm?

dtomalty: Confirmed, lets leave it like this then.

src/lib/thFilter/filters/select/selectFilter.service.coffee
```diff
@@ -10,10 +10,14 @@ angular.module "ThemisComponents"
       @valueField = options.selectOptionsValueField or "value"
 
       if options.selectOptionsUrl?
-        @model = if initialState? then {value: initialState.value} else null
+        @model =
```
dtomalty: On second thought, should this just be `@model = initialState`...

src/lib/thFilter/filters/select/selectFilter.service.coffee
```diff
@@ -10,10 +10,10 @@ angular.module "ThemisComponents"
       @valueField = options.selectOptionsValueField or "value"
 
       if options.selectOptionsUrl?
-        @model = if initialState? then {value: initialState.value} else null
+        @model = initialState
```
dtomalty: Would the behaviour of the select drop down change if `@model` was `undefined` instead of `null`?
My guess is not, but if there was a chance, we should make this `@model = initialState ? null`

PR #338, opened 2017-01-05
CLIO-42843: Disable thButton while loading

craigcarlyle:


src/lib/thActionBar/thActionBarDelegate.service.coffee
```diff
@@ -63,7 +63,10 @@ angular.module "ThemisComponents"
 
       isDisabled = ->
         actionsExist = not not availableActions
-        results.processing or (not results.selectedAction and actionsExist)
+        if availableActions
```
frankieyan: Let's use actionsExist here since it's been casted to a boolean?

PR #340, opened 2017-01-07
Clio 42784 custom filter spurious events fix

JeffMarvin:


src/lib/thFilter/filters/input/thFilter.input.directive.coffee
```diff
@@ -1,20 +1,39 @@
...
+        if newValue isnt lastValue
+          lastValue = newValue
+          return true
+        return false
+
+      @ngKeypress = (event) =>
+        if isUpdatedValue()
+          @filterSet.onFilterChange() if event.which is enterEventCode
+
+      @ngBlur = (event) =>
```
frankieyan: Maybe we should move away from using `ng` prefixes for these ie. `ngBlur`, `ngKeypress`, especially when its no longer exposed in the directive's API. (so something like `onBlur` would be great)

src/lib/thFilter/filters/number/thFilter.number.directive.coffee
```diff
@@ -1,21 +1,36 @@
...
+    controllerAs: "thFilterNumber"
+    template: require "./thFilter.number.template.html"
     controller: ($scope) ->
+      lastValue = null
       enterEventCode = 13
 
+      lastValue = @initialState.value if @initialState?.value?
+
+      isUpdatedValue = =>
+        newValue = @filter.getState()?.value or null
```
frankieyan: Pretty much the same feedback here as thFilter.input

src/lib/thFilter/filters/time/thFilter.time.directive.coffee
```diff
@@ -18,15 +21,24 @@ angular.module "ThemisComponents"
       ]
       @defaultOperatorIndex = 1
 
+      isUpdatedValue = =>
```
frankieyan: This has been reused a few times now, might be a good idea to extract it into its own module or to `utilities.service`?

src/lib/thFilter/filters/input/thFilter.input.directive.coffee
```diff
@@ -1,20 +1,39 @@
...
+      ignoreBlurEvents: "<?"
+      placeholder: "@"
+      initialState: "=?"
     bindToController: true
-    controllerAs: 'thFilterInput'
-    template: require './thFilter.input.template.html'
+    controllerAs: "thFilterInput"
+    template: require "./thFilter.input.template.html"
     controller: ($scope) ->
+      lastValue = ignoreBlurEvents = null
```
frankieyan: The variable `ignoreBlurEvents` doesn't seem to be used. Also, would it be fine leaving `@ignoreBlurEvents` as undefined when it's not provided instead of setting it to null?

src/lib/thFilter/filters/input/thFilter.input.directive.coffee
```diff
@@ -1,20 +1,39 @@
...
+
+      lastValue = @initialState.value if @initialState?.value?
+      @ignoreBlurEvents or= false
```
frankieyan: We'll probably take care of this in the 1.6 update branch, but do keep in mind that these bindings would not be available before the `$onInit` hook is called in the future.

src/lib/thFilter/filters/number/thFilter.number.directive.coffee
```diff
@@ -1,21 +1,36 @@
-angular.module 'ThemisComponents'
-  .directive 'thFilterNumber', (NumberFilter) ->
-    restrict: 'E'
+angular.module "ThemisComponents"
+  .directive "thFilterNumber", (NumberFilter) ->
+    restrict: "E"
     scope:
-      filterSet: '='
-      filterOptions: '='
-      ngBlur: '&'
```
frankieyan: This binding is used in `thCustomFilterRow.template.html` as well

src/lib/thFilter/filters/input/thFilter.input.directive.coffee
JeffMarvin: Is there a better place to put them?

frankieyan: Yep in the `$onInit` method on the controller. Before this hook kicks in any bindings will be undefined.

JeffMarvin: Ok, good call. I was looking for some docs on these lifecycle hooks on *directives*, but couldn't find anything. Any ideas on where that is described?

src/lib/thFilter/filters/input/thFilter.input.directive.coffee
```diff
@@ -1,20 +1,39 @@
...
+
       enterEventCode = 13
 
-      @ngKeypress = (event) ->
-        @filterSet.onFilterChange() if event.which is enterEventCode
+      isUpdatedValue = =>
+        newValue = @filter.getState()?.value or null
```
JeffMarvin: `@filter.getState()` can return null in some cases, and undefined in others. 
So, I'm trying to make sure the comparison (`if newValue isnt lastValue`) works out. Thoughts?

src/lib/thFilter/filters/input/thFilter.input.directive.coffee
JeffMarvin: Good idea, I've moved these to use an "on" prefix: `onBlur` and `onKeypress`.

src/lib/thFilter/filters/number/thFilter.number.directive.coffee
JeffMarvin: Caught it, thanks.

src/lib/thFilter/filters/time/thFilter.time.directive.coffee
JeffMarvin: I like that idea. I am currently using this method to also modify `lastValue` as well as return a boolean. Thoughts on how to structure that method if I put it in a separate service?

src/lib/thFilter/filters/input/thFilter.input.directive.coffee
frankieyan: Yep -> https://docs.angularjs.org/api/ng/service/$compile

src/lib/thFilter/filters/input/thFilter.input.directive.coffee
```diff
@@ -35,6 +31,11 @@ angular.module "ThemisComponents"
           unless @ignoreBlurEvents
             @filterSet.onFilterChange()
 
+      @$onInit = ->
+        lastValue = @initialState.value if @initialState?.value?
+        @ignoreBlurEvents or= false
+        return
```
frankieyan: Don't really need a return statement here, there is nothing expecting this output anyway

src/lib/thFilter/filters/time/thFilter.time.directive.coffee
frankieyan: Discussed - no need to extract 👍

src/lib/thFilter/filters/number/thFilter.number.directive.coffee
```diff
@@ -1,21 +1,36 @@
...
     bindToController: true
-    controllerAs: 'thFilterNumber'
-    template: require './thFilter.number.template.html'
+    controllerAs: "thFilterNumber"
+    template: require "./thFilter.number.template.html"
     controller: ($scope) ->
+      lastValue = undefined
       enterEventCode = 13
 
+      lastValue = @initialState.value if @initialState?.value?
```
frankieyan: I might've forgot to point this out before, but this binding usage should also be in `$onInit` like `thFilter.input.directive.coffee` above. Same with `thFilter.time` below.

src/lib/thFilter/filters/time/thFilter.time.directive.coffee
```diff
@@ -10,6 +10,8 @@ angular.module "ThemisComponents"
     controllerAs: "thFilterTime"
     template: require "./thFilter.time.template.html"
     controller: ($scope, $element) ->
+      lastValue = @initialState.value if @initialState?.value?
```
frankieyan: `lastValue = @initialState?.value` should get us what we want yeah? (Also move this into `$onInit`)

src/lib/thFilter/filters/time/thFilter.time.directive.coffee
```diff
@@ -18,22 +19,34 @@ angular.module "ThemisComponents"
...
+          @filterSet.onFilterChange()
 
       @onKeypress = (event) ->
         @validateInput() if event.which is enterEventCode
 
       @onOperatorChange = ->
         @filterSet.onFilterChange() if @filter.time
 
+      @$onInit = ->
+        lastValue = @initialState.value?
```
frankieyan: Should be `@initialState?.value` right? Since we want to check for `@initialState`'s existence before pulling `value` out

src/lib/thFilter/filters/time/thFilter.time.directive.coffee
```diff
@@ -20,7 +20,7 @@ angular.module "ThemisComponents"
       @defaultOperatorIndex = 1
 
       isUpdatedValue = =>
-        newValue = @filter.model
+        newValue = @filter.model || undefined
```
frankieyan: How about `newValue = @filter?.model`?

JeffMarvin: yyyup

JeffMarvin: Actually, here's why I was doing the `|| undefined`
`@filter.model` becomes `""` when the field is cleared, but is `null` when the field is first initialized. So, I was using this line to "normalize" that so that the comparison between new and last value worked.

PR #342, opened 2017-01-10
CLIO-43152: Ensure that thSelect selected attributes get cleared when option is changed

craigcarlyle:
* Ensured thSelect `option` elements get the `selected` attribute cleared when the option is changed
* Made styling changes in `src/lib/thSelect/thSelect.directive.coffee`
  * Replaced single quotes with double quotes
  * Replaced `&&` with `and`


src/lib/thSelect/thSelect.directive.coffee
```diff
@@ -57,7 +57,15 @@ angular.module('ThemisComponents')
           if isSelected element
             @ngModel = {name: element.text, value: element.value}
 
-      @thOnChange = -> Utilities.onChange @ngChange
+      @thOnChange = ->
+        Utilities.onChange @ngChange
+
+        # Clear any other selected options
+        optionElements = $element.find "option"
+        for optionElement in optionElements
```
frankieyan: Can we use a foreach here instead: https://github.com/slapheadted/themis/blob/coffeescript-style-guide/dev-docs/coding_style_guide/coffeescript_style_guide.md#favour-native-functions-over-comprehensions

Also I believe 1.5.10/1.6 is doing something with the selected attribute on `option` elements so we might need to revisit this or potentially take this out when upgrading.

craigcarlyle: Addressed.

PR #345, opened 2017-01-11
CLIO-43323 - Removing promise rejection when modal is dismissed

jhnbrnn:
Angular 1.6 throws an error whenever a rejected promise is not handled. During the upgrade process, we noticed these errors being thrown when modals were dismissed. From our investigation, we don't ever handle "dismiss modal" cases in Themis, so for simplicity's sake (rather than disabling the error across the site) we've removed the promise rejection from the dismiss call in the ModalManager.


jhnbrnn:
closed in favour of #346

PR #347, opened 2017-01-13
Replaced express with browsersync

mbuckley:
This update introduces two changes to ThemisUI.
- express replaced with browsersync
- TypeScript loader added

The distribution method and build process has not changed, however we will still want to test the integration with Themis


src/docs-app/catalog/bind-markdown/examples/1 - Simple/coffee.coffee
```diff
@@ -1,5 +1,5 @@
-angular.module('thDemo', ['ThemisComponents'])
-  .controller 'DemoController', ($scope) ->
+angular.module("docsBindMarkdownDemo")
```
frankieyan: I'm thinking examples for doc components arent needed anymore

mbuckley: 👍  - agree.

src/docs-app/catalog/catalog.service.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  /* @ngInject */
+  constructor() {
+    this.version = componentCatalog.version;
+    this.components = componentCatalog.components;
+    this.docs = componentCatalog.docs;
+  }
+
+  getComponent(name: string) {
+    return this.components.filter(component => component.name === name)[0];
```
frankieyan: Not necessarily now but we should find a TS compatible set of polyfills here and use `.find`. (https://github.com/zloirock/core-js ? Used by Angular 2)

src/docs-app/catalog/component-example/component-example.template.html
```diff
@@ -0,0 +1,37 @@
...
+      scrolling="no"
+      >
+    </iframe>
+    <pre ng-show="mode === 'html'" class="line-numbers"><code class="language-markup">{{example.html}}</code></pre>
+    <pre
+      class="line-numbers"
+      ng-show="mode === key"
+      ng-repeat="(key, value) in example.others track by key"
+      ><code ng-class="getLanguage(key)">{{value}}</code></pre>
+    <pre ng-show="mode === 'coffee'" class="line-numbers"><code class="language-coffeescript">{{example.coffee}}</code></pre>
```
frankieyan: We should add ts examples support here, perhaps a follow up ticket?

src/docs-app/catalog/component-example/index.scss
```diff
@@ -27,7 +26,7 @@ docs-component-example,
       display: inline-block;
       padding: 0rem 2rem;
 
-      color: $white;
+      color: #fff;
```
frankieyan: This should be `$white`

src/docs-app/catalog/component-list/component-list.template.html
```diff
@@ -0,0 +1,33 @@
+<div class="sidebar">
+  <header class="main">
+    <!-- <i class="fa fa-home" aria-hidden="true"></i> -->
```
frankieyan: Remove if unused

src/docs-app/docs-app/docs-app.scss
```diff
@@ -87,8 +91,8 @@ h1, h2, h3 {
     text-align: center;
     text-decoration: none;
 
-    border-top: $border;
-    border-right: $border;
+    border-top: solid 1px $grey-4;
```
frankieyan: These should be `$border`

src/docs-app/docs-app/docs-app.ts
```diff
@@ -0,0 +1,18 @@
...
+class DocsApp {
+
+  /* @ngInject */
+  constructor() {}
+
+}
+
+angular.module("ThemisComponentsApp")
+  .component("docsApp", {
+    controller: DocsApp,
```
frankieyan: We can remove the `DocsApp` class and the controller property here if it's not needed.

src/docs-app/docs-app/docs-app.ts
```diff
@@ -0,0 +1,18 @@
+import * as angular from "angular";
+
+require("./docs-app.scss");
+
+const template = require("./docs-app.template.html") as string;
+
+class DocsApp {
```
frankieyan: Remove

src/docs-app/routes.coffee
```diff
@@ -0,0 +1,31 @@
...
+        name: "index"
+        url: "/"
+        component: "docsComponentDetails"
+
+      .state
+        name: "component"
+        url: "/component/:name"
+        component: "docsComponentDetails",
+        resolve:
+          # name: ($transition$) -> $transition$.params().name
```
frankieyan: This comment was not-working code and can be removed

src/docs-app/catalog/component-example/component-example.directive.coffee
```diff
@@ -0,0 +1,124 @@
...
+        "#{includeBase}assets/docs-vendor.js",
+        "#{includeBase}assets/examples.js"
+      ]
+      includeCSS = [
+        "#{includeBase}assets/examples.css",
+        "#{includeBase}assets/lib-themisui-styles.css"
+        "//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css"
+      ]
+      configBlock = """
+        .config(($sceDelegateProvider) ->
```
frankieyan: @JefCurtis We added this to white list localhost with $sce when serving examples over codepen. This is only needed with components that use ng-include, and has a flag set on the individual components' meta.json

src/docs-app/catalog/component-example/component-example.directive.coffee
```diff
@@ -0,0 +1,124 @@
...
+          $sceDelegateProvider.resourceUrlWhitelist [
+            "self"
+            "http://localhost:3042/**"
+            "http://0.0.0.0:3042/**"
+          ]
+        )
+      """
+      penExampleCoffee = example.coffee
+        .replace moduleRegex, (match) -> match + if $scope.whitelistLocal then configBlock else ""
+        .replace externalFilePathRegex, prependHost
```
frankieyan: @JefCurtis We've also made a change here to make the external file paths absolute when serving over codepen (thModal etc)

src/docs-app/routes.coffee
```diff
@@ -0,0 +1,31 @@
+angular.module("ThemisComponentsApp")
+  .config ($stateProvider, $locationProvider, $urlRouterProvider) ->
+    $locationProvider.html5Mode true
+
+    $stateProvider
```
frankieyan: @JefCurtis this is new - the routing for the components are now done with ui-router

src/lib/thPopover/examples/1 - Simple/html.html
```diff
@@ -1,11 +1,11 @@
-<html ng-app="thDemo">
+<html ng-app="thPopoverDemo">
   <body style="height: 200px;">
 
-    <th-button th-popover-url="exampleTemplates/thPopoverExampleTemplate.html">Trigger</th-button>
-    <th-button th-popover-url="exampleTemplates/thPopoverExampleTemplate2.html">Trigger 2</th-button>
-    <th-button th-popover-url="exampleTemplates/thPopoverExampleTemplateTiny.html">Tiny</th-button>
-    <th-button th-popover-url="exampleTemplates/thPopoverExampleTemplate.html" style="float: right">Right</th-button>
-    <th-button th-popover-url="exampleTemplates/thPopoverExampleTemplate2.html" style="float: right">Right 2</th-button>
+    <th-button th-popover-url="/components/thPopover/examples/1 - Simple/thPopoverExampleTemplate.html">Trigger</th-button>
```
frankieyan: @JefCurtis component examples' external files are now inline with the components. These are the urls that get converted to absolute paths in codepen.

frankieyan: Hm is there a strong use case for adding css in examples? We can always use a style tag for these snippets.

src/docs-app/catalog/bind-markdown/readme.md
```diff
@@ -1,6 +1,6 @@
-# `bindMarkdown`
+# `thBindMarkdown`
 
-The `bindMarkdown` component evaluates a Markdown string and binds it to the contents of the given element.
+The `docsBindMarkdown` component evaluates a Markdown string and binds it to the contents of the given element.
```
frankieyan: To clarify that its used for the doc app only, we're not bundling this into the ThemisUI dist anymore

src/docs-app/catalog/component-details/component-details.component.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+angular.module("ThemisComponentsApp")
+  .component("docsComponentDetails", {
+    bindings: {
+      name: "<",
+      component: "<"
+    },
+    controller: ComponentDetails,
+    template
```
frankieyan: Yup. Same thing in coffeescript I believe.

src/docs-app/catalog/component-example/component-example.directive.coffee
```diff
@@ -0,0 +1,124 @@
...
+            "<script src=\"#{scriptSrc}\"></script>"
+          html = html.replace "</body>", "#{scripts.join("")}</body>"
+
+          exampleFrame.contentWindow.document.open()
+          exampleFrame.contentWindow.document.write html
+          exampleFrame.contentWindow.document.close()
+          observer = new MutationObserver (mutations) ->
+            # Just accessing scrollTop seems to fix a weird FF flicker error.
+            # Once FF supports `scrollingElement` we can take out our polyfill.
+            # CLIO-37920
```
frankieyan: Yeah we'll be creating follow up tickets to convert these to TS, maybe this can be a part of that work.

src/docs-app/catalog/component-example/component-example.directive.coffee
```diff
@@ -0,0 +1,124 @@
...
+            exampleFrame.style.height = exampleFrame
+                                          .contentWindow
+                                          .document
+                                          .body
+                                          .scrollHeight + 10 + "px"
+          , 1500
+
+          $timeout -> Prism.highlightAll()
+
+      $scope.getLanguage = (fileName) ->
```
frankieyan: In a follow up 😅

src/docs-app/catalog/catalog.service.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  getDoc(name: string) {
+    return this.docs.filter(doc => doc.name === name)[0];
+  }
+
+  parseComponentName(camelCase: string): string {
```
frankieyan: Hm I think `:string` here can be removed as TS should be able to infer that we're returning a string in all cases here.

src/docs-app/catalog/component-details/component-details.component.ts
```diff
@@ -0,0 +1,56 @@
...
+    };
+
+    this.examples = this.component.examples;
+    this.private = this.component.private;
+    this.whitelistLocal = this.component.whitelistLocal;
+  }
+
+  exampleWhitelistLocal(exampleIndex: number) {
+    if (this.whitelistLocal instanceof Array) {
+      return this.whitelistLocal.some(whitelisted => whitelisted === exampleIndex + 1);
```
frankieyan: I believe one is using a 0-based index and the other a 1-based index.

src/docs-app/catalog/catalog.interfaces.ts
```diff
@@ -0,0 +1,50 @@
+interface Dictionary {
+  [filename: string]: string;
```
frankieyan: It still enforces that each property in an object of this type only holds a string.

src/docs-app/catalog/catalog.service.ts
```diff
@@ -0,0 +1,30 @@
+import * as angular from "angular";
+import {Catalog, Component} from "./catalog.interfaces";
+const componentCatalog = require("./catalog") as Catalog;
+
+export class CatalogService {
+  version: string;
+  components: Component[];
+  docs: Component[];
+
+  /* @ngInject */
```
frankieyan: Will double check the output but our current coffeescript comment outputs this:
```js
/*@ngInject */
CustomFilters.$inject = ["CustomFilterConverter", "$http", "$timeout"];
```

src/docs-app/catalog/component-details/component-details.component.ts
frankieyan: So yes I think creating a variable here to hold the +1'ed index is a good idea, something like `const normalizedIndex = exampleIndex + 1`?

src/docs-app/catalog/component-example/component-examples.scss
```diff
@@ -1,8 +1,6 @@
-@import "../../themes/themis/index";
-@import "../../lib/index";
```
frankieyan: Moved over here: https://github.com/mbuckley/lib-themisui/blob/c0c7b547770d3e3968073c9032ad472f07d9f0e0/src/docs-app/docs-app/docs-app.ts#L3

src/docs-app/catalog/component-example/index.scss
```diff
@@ -37,13 +36,6 @@ docs-component-example,
       cursor: pointer
     }
 
-    input[type="submit"] {
```
frankieyan: Not too sure, might've been some button we used to have on the example blocks. The button below is for the codepen button though, and we have no other submit input buttons.

src/docs-app/catalog/component-list/component-list.component.ts
```diff
@@ -0,0 +1,25 @@
+import * as angular from "angular";
+import {CatalogService} from "../catalog.service";
```
frankieyan: Not sure if typescript can destructure into object properties instead of variables, have to see.

src/docs-app/catalog/component-list/component-list.template.html
```diff
@@ -0,0 +1,33 @@
...
+    <h1>
```
frankieyan: Oh I dunno, don't feel strongly either way. We can make the change tomorrow.

src/docs-app/documentation/colourPalette/colourPalette.component.ts
```diff
@@ -0,0 +1,35 @@
+import * as angular from "angular";
+type scssVariables = {
+  [id: string]: string;
+};
+// tslint:disable-next-line:max-line-length
+const scssVariables = require("!!sass-variable-loader?preserveVariableNames!../../../themes/themis/colors.scss") as scssVariables;
```
frankieyan: It's to tell webpack to ignore the preconfigured loaders for scss files here, meaning we don't actually want this to be a part of our stylesheet here, and instead only use the sass-variable-loader for this specific usage.

src/docs-app/documentation/colourPalette/colourPalette.component.ts
```diff
@@ -0,0 +1,35 @@
...
+    hex: string;
+  }[];
+
+  /* @ngInject */
+  constructor() {}
+
+  $onInit() {
+    this.colours = Object.keys(scssVariables)
+      .filter(variable => scssVariables[variable].search(/^#([a-f0-9]{3,6})/i) >= 0)
+      .sort()
```
frankieyan: Yeah unicode, not sure if we really need to here, we can always ref docs.

src/docs-app/documentation/colourPalette/index.scss
```diff
@@ -0,0 +1,26 @@
+.colour-palette-container {
```
frankieyan: Hah maybe...

src/lib/thActionBar/examples/1 - Simple/coffee.coffee
```diff
@@ -1,5 +1,5 @@
-angular.module 'thDemo', ['ThemisComponents']
-  .controller "DemoController", (
+angular.module("thActionBarDemo")
+  .controller "thActionBarDemoCtrl1", (
```
frankieyan: This was done to support importing modules in the examples (index files are generated with a script, but git-ignored to tie the require tree together). But looking at it again it doesn't work on codepen.

src/docs-app/catalog/component-list/component-list.component.ts
frankieyan: `{ this.a, this.b } = test;` isn't valid JS/TS, you're only able to do: `const { a, b } = test;`

src/docs-app/index.html
```diff
@@ -3,66 +3,15 @@
   <head>
     <meta charset="utf-8">
     <title>ThemisUI</title>
-    <style type="text/css">
-      [ng\:cloak], [ng-cloak], [data-ng-cloak], [x-ng-cloak], .ng-cloak, .x-ng-cloak {display: none !important;}
```
frankieyan: No, we're not really seeing flashes of unbinded data in the docs app. If we do then one solution would be to load docs-vendor.js in the head and have angular inject these styles for us.

src/docs-app/routes.coffee
```diff
@@ -0,0 +1,31 @@
...
+      .state
+        name: "index"
+        url: "/"
+        component: "docsComponentDetails"
+
+      .state
+        name: "component"
+        url: "/component/:name"
```
frankieyan: ??

src/docs-app/catalog/catalog.service.ts
frankieyan: ^ Works

src/lib/thPopover/examples/1 - Simple/html.html
JefCurtis: Yep we are already using CSS script tags. Would be nice to be a able to pull it out and keep the examples as small as possible though.

src/docs-app/catalog/component-example/component-example.directive.coffee
JefCurtis: Cool. I'll assume that this can be taken care of with the follow up ticket.

src/docs-app/documentation/colourPalette/colourPalette.component.ts
JefCurtis: Feel like we should either remove the `sort` or explicitly sort by `name` or something.

src/docs-app/routes.coffee
JefCurtis: Should the `.state` function expect a string to be used here. We were having problems in Themis where the ui-router was trying to parse a `string` value into a id integer and failing hard. wondering if this could be a problem with ThemisUI if we were planning to add query values to the url.

src/lib/thPopover/examples/1 - Simple/html.html
frankieyan: Next steps!

src/docs-app/catalog/catalog.service.ts
frankieyan: Looking back it might be a good idea to always annotate the return type of a function, we can discuss around this once we all sink in on TS a little bit and then choose a styleguide.

src/docs-app/catalog/component-details/component-details.component.ts
frankieyan: Moving the `+ 1` to the template, so this function would always expect a 1-based index that the examples use.

src/docs-app/catalog/component-list/component-list.template.html
frankieyan: Pre-HTML5 it was a strict requirement to not wrap block elements with inline elements - still valid today just that the other way is also allowed now. Don't think this is a hard-requirement either way so let's not change it here, unless we need the whole header block to be clickable in the Apollo docs redesign.

src/docs-app/documentation/colourPalette/colourPalette.component.ts
frankieyan: But these are just strings we're sorting here, and the default usage of array.sort gives us an alphabetical ordered list.

src/docs-app/documentation/colourPalette/index.scss
frankieyan: Discussed -> Not the right usage for flex here.

src/docs-app/routes.coffee
frankieyan: Looking at the v1 docs it looks like the urlmatcher always returns these fragments as strings, even if it was a number. Might be due to the different version of ui-router being used, but we can always debug when we need to start adding more routes to the docs.

PR #348, opened 2017-01-15
Clio 41708 colour palette doc

frankieyan:


src/docs-app/documentation/colourPalette/colourPalette.template.html
```diff
@@ -1,12 +1,18 @@
-<h1>Colour</h1>
+<h1>Colours</h1>
```
mbuckley: Let's use American Color here..it will be confusing because the css needs to be merica.

PR #357, opened 2017-01-19
Clio 42851 downgrade jquery

frankieyan:
Originally wanted to exclude jQuery from the dist build by declaring it as external, with webpack pulling it in from the global context. This isn't possible though when our library target is `commonjs` or `commonjs2`. The given string will be used to resolve to `require("jQuery")` which will cause parent projects (with browserify or webpack) to look for the given string as a module name under `node_modules`.


jhnbrnn:
Just want to make sure I've got this correctly: with this change, ThemisUI will still include jQuery in the `dist` build, but it downgrades to the same version Themis currently relies on. The other changes (`component-example` and the `docsServer` gulp task) just ensure jQuery is included in the docs server when it runs?

frankieyan:
Yep unfortunately it's still gonna be packed in. The changes to component example loads jQuery ahead of Angular so Angular would pick that up instead of using jQLite.

jhnbrnn:
Cool, makes sense. Looks good to me!

PR #362, opened 2017-01-30
Clio 43468 th modal over th modal

JeffMarvin:
In this PR, I make a small change to the `thModalAnchor.template.html` and otherwise just construct a new Example 8 to show the Modal over Modal behaviour.

Navigate to Example 8 and observe how the first modal contains a button to edit the email address. This button launches a second modal over the first. It returns its data down properly to the originating modal, and then to the page once the modal has submitted.

This is functional only, styling of the appearance will happen in a future ticket.


jhnbrnn:
Overall, I think this is looking pretty good. I've got one recommendation that touches a few areas of the code, so I'll put the main point here & annotate the specific files afterward:

We probably want to indicate, in a programmatic way, whether a modal is a full-screen or a "standard" modal. To do that, I think we'd need a few things:
* ModalManager would need to set a "fullscreen" (or similarly named) property on the modal object it stores in its `_modals` Array
* The modal itself should probably bind the "fullscreen" property to a local variable (which could later be used to differentiate styling, but that's for another ticket)
* the `thModalAnchor` should check the modal object and set the `limitTo` property on `ng-repeat` to either 1 or 2 based on whether or not it's a full-screen modal - that way, "standard" modals won't be able to launch other modals.

JeffMarvin:
I will make these suggestions as part of this branch. The intent is for this branch to add basic "fullpage" modal functionality, so it makes sense.

JeffMarvin:
I pushed some new stuff up. Tests need to be fixed, and so does thModalAnchor.template.html but this stubs in the basic awareness of the "type" of modal.

JeffMarvin:
Passing tests, and added in a datepicker in the top level modal as a demo that the th-popover is working as expected.
Also have implemented the suggestions in your comment above.
I will circle back on the requirements from the design doc, as there are some new ones not identified when the ticket was created. I will recommend that we tackle some of the new behaviors under a new ticket/PR.

src/lib/thModal/thModalAnchor.template.html
```diff
@@ -1,6 +1,13 @@
...
-    <div class="cell">
+  <pre>{{anchor.modals|json}}</pre>
+  <div
+    class="th-modal-container"
+    ng-repeat="modal in anchor.modals | limitTo:anchor.modalLimit"
+    >
+    <div ng-if="modal.type == 'fullpage'" class="cell">
+      <th-modal modal-data="modal"></th-modal>
+    </div>
+    <div ng-if="modal.type == 'dialog'" class="cell">
```
jhnbrnn: the `cell` div here doesn't seem to do anything different whether the modal is of type `dialog` or `fullpage`. Is the intent to eventually differentiate the two?

src/lib/thModal/thModalManager.service.coffee
```diff
@@ -9,31 +9,33 @@ angular.module("ThemisComponents")
...
       deferred = $q.defer()
       modalPromise =
         if template?
-          $q.when(data: template)
+          $q.when data: template
         else
-          $http(url: path, method: "GET", params: params)
-
+          $http url: path, method: "GET", params: params
+
```
jhnbrnn: The tiniest nitpick, but some extra whitespace got added on this line.

jhnbrnn:
Looks good! I've got one question about the `thModalAnchor` template, but other than that, this looks good to ship. Agree with using a new ticket/PR for the new behaviours.

Is there anything we should be unit testing about the fullpage thModal, or do you think those will come later if it starts to noticeably differ in functionality from the dialog thModal?

src/lib/thModal/thModalAnchor.template.html
JeffMarvin: I opted to keep this simple and symbolic for now. The next ticket will style this, and depending on the CSS/HTML structure, this template logic could change considerably.

src/lib/thModal/thModalManager.service.coffee
JeffMarvin: Fixing.

JeffMarvin:
Closing this in order to open a new PR against a new branch that will be used to gather the work under the JIRA Epic [CLIO-43451](https://themis.atlassian.net/browse/CLIO-43451).

PR #365, opened 2017-01-30
CLIO-43318: Kendo Autocomplete

craigcarlyle:


src/docs-app/catalog/component-example/component-example.directive.coffee
```diff
@@ -31,9 +31,9 @@ angular.module("ThemisComponentsApp")
           ]
         )
       """
-      penExampleCoffee = example.coffee
-        .replace moduleRegex, (match) -> match + if $scope.whitelistLocal then configBlock else ""
-        .replace externalFilePathRegex, prependHost
+      #penExampleCoffee = example.coffee
```
craigcarlyle: Disabled these temporarily as they were causing errors in `ts` files.

src/lib/thAutocomplete/autocomplete.component.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (this.name)
+      $element.find("input").attr("name", this.name);
+
+    if (this.condensed)
+      $element.attr("condensed", this.condensed);
+  }
+
+  public $onChanges(change: any) {
+    setTimeout(() => {
+      if (change.ngDisabled && typeof(change.ngDisabled.previousValue) === "boolean")
```
craigcarlyle: Checking if `previousValue` is a boolean because a change event is called upon initialization and the initial value is blank. Wondering if there's a more elegant way of ensuring these events don't get called immediately.

src/lib/thAutocomplete/autocomplete.service.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  public create(options: any) {
+    if (options["ngDisabled"]) {
+      this.enabled = false;
+    }
+
+    if (options.combobox) {
+      new kendo.ui.ComboBox(options["element"], {
+        dataTextField: options["delegate"]["dataTextField"] ? options["delegate"]["dataTextField"] : "name",
```
craigcarlyle: These are greater than 100 characters, but I'm not sure how we want to format multiple-line ternary operations in TS.

package.json
```diff
@@ -25,6 +25,7 @@
     "npm": "3.8.8"
   },
   "dependencies": {
+    "@progress/kendo-ui": "^2017.1.118",
```
frankieyan: Install `@types/kendo-ui` here as well. Use the --save-exact flag so we shrinkwrap the exact version (for `@progress/kendo-ui` too)

src/lib/index.coffee
```diff
@@ -1,3 +1,7 @@
+$ = require 'jquery'
+angular = require 'angular'
+require 'kendo-ui-core'
```
frankieyan: Update to `@progress/kendo-ui`

src/lib/services/data-source.service.coffee
```diff
@@ -0,0 +1,17 @@
+angular = require "angular"
+require "kendo-ui-core/js/kendo.data.js"
```
frankieyan: Update to `@progress/kendo-ui/js/kendo.data` (should be able to omit the extension)

src/lib/thAutocomplete/autocomplete.service.ts
```diff
@@ -0,0 +1,102 @@
+import * as $ from "jquery";
+import angular = require("angular");
```
frankieyan: This section can be:

```typescript
import * as angular from "angular";
import "@progress/kendo-ui/js/kendo.autocomplete";
import "@progress/kendo-ui/js/kendo.combobox";
import "@progress/kendo-ui/js/kendo.multiselect";
```

We can also remove the global declare below by adding the kendo-ui type to tsconfig.json

src/lib/thAutocomplete/autocomplete.service.ts
```diff
@@ -0,0 +1,102 @@
...
+  public create(options: any) {
+    if (options["ngDisabled"]) {
+      this.enabled = false;
+    }
+
+    if (options.combobox) {
+      new kendo.ui.ComboBox(options["element"], {
+        dataTextField: options["delegate"]["dataTextField"] ? options["delegate"]["dataTextField"] : "name",
+        dataValueField: options["delegate"]["dataValueField"] ? options["delegate"]["dataValueField"] : "id",
+        enabled: this.enabled,
```
frankieyan: Kendo-ui's type definition caught this -> Should be `enable` (http://docs.telerik.com/kendo-ui/api/javascript/ui/autocomplete#configuration-enable)

tsconfig.json
```diff
@@ -8,7 +8,8 @@
     ],
     "experimentalDecorators": true,
     "noUnusedLocals": true,
-    "noUnusedParameters": true
+    "noUnusedParameters": true,
```
frankieyan: Once `@types/kendo-ui` is installed, add it to the global types array above:
```js
    "types": [
      "webpack-env", "kendo-ui"
    ],
```

src/lib/thAutocomplete/autocomplete.service.ts
```diff
@@ -0,0 +1,102 @@
...
+    this.noDataTemplate = "No results.";
+  }
+
+  public create(options: any) {
+    if (options["ngDisabled"]) {
+      this.enabled = false;
+    }
+
+    if (options.combobox) {
+      new kendo.ui.ComboBox(options["element"], {
```
frankieyan: How come we are using bracket notations through out this file here?

PR #366, opened 2017-01-31
Webpack 2.2 + jQuery 1.12

frankieyan:


jhnbrnn:
You might need to run `npm shrinkwrap` one more time, I don't think the webpack 2.2 upgrade was captured in the shrinkwrap file on my branch.

PR #370, opened 2017-02-01
Added ng-strict-di and updated port settings to work with heroku

mbuckley:


tools/gulp/Config.ts
```diff
@@ -3,9 +3,36 @@ const historyApiFallback = require("connect-history-api-fallback");
...
+
+if(process.env.PORT) {
+  ui = false;
+  ghostMode = false;
+}
+
 const ConfigObj = {
   browserSync: {
-    port: 3042,
+    port: process.env.PORT ? process.env.PORT : 3042,
```
frankieyan: How about `process.env.PORT || 3042`

PR #383, opened 2017-02-10
Clio 44270 thSwitch improvements

Luxiyalu:
https://themis.atlassian.net/browse/CLIO-44270

Previous thSwitch not triggering the digest cycle outside of element properly. Note: using ng-click on the span, so that the compile can be processed by the angular system. Since we're using an invisible <input>, for the problem of ng-change cannot be triggered programmatically, we will manually call ngChange when toggle is called.


src/lib/thSwitch/thSwitch.directive.coffee
```diff
@@ -5,7 +5,7 @@ angular.module('ThemisComponents')
     template: require('./thSwitch.template.html')
     scope:
       name: '@'
-      change: '&ngChange'
+      ngChange: '&'
```
JefCurtis: Since you are already in this file we should clean it up to match our style guide. Mostly remove unneeded `()` and use double quotes over single quotes. 
https://github.com/clio/themis/blob/master/dev-docs/coding_style_guide/coffeescript_style_guide.md

src/lib/thSwitch/thSwitch.directive.coffee
```diff
@@ -14,11 +14,7 @@ angular.module('ThemisComponents')
       @state = @state ? off
```
JefCurtis: Unused injections `$scope` and `$element`.

JefCurtis:
I suspect the tests might need to be updated as well.

Luxiyalu:
@JefCurtis yes, I'll update the tests, and do it with style.

